PR #1, opened 2019-10-02
FIX glob pattern in windows

miyamoen:
Thank you for a great tool🎉

I was happy and tried it right away! 

image

there was src/Main.elm...

In windows, ["glob" require only forward-slashes](https://github.com/isaacs/node-glob#windows)

fixed it, then:

image

Could you fix it like this?

### DIFF
--- a/lib/elm-files.js
+++ b/lib/elm-files.js
@@ -137,7 +137,7 @@ function getFilesToReview(
 
 function findFiles(directory) {
   return {
-    files: glob.sync(directory + '/**/*.elm', {
+    files: glob.sync((directory.startsWith('C:') ? directory.slice(2).replace(/\\/g, '/') : directory) + '/**/*.elm', {
       nocase: true,
       ignore,
       nodir: false
--- a/lib/elm-files.js
+++ b/lib/elm-files.js
@@ -138,9 +138,7 @@ function getFilesToReview(
 function findFiles(directory) {
   return {
     files: glob.sync(
-      (directory.startsWith('C:')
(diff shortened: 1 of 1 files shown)


jfmengels:
Hi @miyamoen!

First of all, thank you for your interest and taking the time to find how to fix it :)

From what I read in the `glob` documentation, [backslash characters should not be used](https://www.npmjs.com/package/glob#windows), so your fix looks to me like it goes in the right direction.

That said, I think the suggested fix will not work for all cases, because potentially, other users will have their code under `D:`, and not `C:`.

I think the following code would do the trick. Could you try that out and see how that works well for you?
```elm
files: glob.sync(directory.replace(/\\/g, '/') + '/**/*.elm', {
```

miyamoen:
Thanks for reply🙂

tried this code:
`files: glob.sync(directory.replace(/\\/g, '/') + '/**/*.elm', {`

image

did not work well!

in the `glob` documentation
> On windows, this will by default result in /foo/* matching C:\foo\bar.txt.
I don't know why.


` files: glob.sync(directory.replace(/.:/, "").replace(/\\/g, "/") + "/**/*.elm", {`
This code works well.

jfmengels:
Ok. That seems to still work fine on my Linux computer. If it works on yours, let's change the line to your suggestion :+1: 

Can you make the edit, or do you prefer I take over?

miyamoen:
Edited!

jfmengels:
Thank you @miyamoen!

I have published the fix in the 1.0.1 release. I hope you'll find the tool useful (and usable)!

PR #11, opened 2020-05-27
Use --stdin option of elm-format

sparksp:
This saves having to write the file to disk before immediately asking elm-format to load it and change it. This may save a little processing time and should prevent any double hits for other tools watching the files.


jfmengels:
Seems to work fine :+1: 
Thanks a lot :blush:

PR #23, opened 2020-10-15
Make report async

Arkham:
Hello there! I've bumped into the following issue.

With this setup:
```
$ node --version
v13.8.0
$ npm --version
6.13.6
$ npx elm-review --version
2.3.0
```

Running the following commands:
```
$ git clone git@github.com:jfmengels/elm-review-unused.git
$ cd elm-review-unused
$ npm install
$ npm test
```

Led to this error:

```
An error occurred while trying to check whether the preview configuration compiles.
Error: Command failed: npx elm-review --config /Users/arkham/code/elm-review-unused-test/preview --report=json
    at checkExecSyncError (child_process.js:611:11)
    at execSync (child_process.js:647:15)
    at checkThatExampleCompiles (/Users/arkham/code/elm-review-unused-test/elm-review-package-tests/check-previews-compile.js:16:5)
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (/Users/arkham/code/elm-review-unused-test/elm-review-package-tests/check-previews-compile.js:12:29)
    at Module._compile (internal/modules/cjs/loader.js:1151:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
```

Which was caused by the output of elm-review being cut at 8192 characters.

Doing some digging this seems to be a common issue in NodeJS when doing something like:
```
console.log(veryLargeString);
process.exit();
```
Check out these links for more info:
- https://nodejs.org/api/process.html#process_process_exit_code
- https://github.com/nodejs/node/issues/6456
- https://github.com/nodejs/node/issues/19218

So in this PR I changed the `report` function to be async and replaced `console.log` with `process.stdout.write`, which allows to be passed a callback when the write has been finished (https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback). I also promisified it so we don't have to pass that callback from the normal code.

With this fix in place, running `npm test` in the above setup succeeds.

I have no idea about NodeJS development in general, so please let me know if this looks good to you.

### DIFF
--- a/lib/report.js
+++ b/lib/report.js
@@ -2,7 +2,7 @@ const styledMessage = require('./styled-message');
 
 module.exports = report;
 
-function report(options, result) {
+async function report(options, result) {
   if (options.report === 'json') {
     return print(options, jsonReport(result.errors));
   }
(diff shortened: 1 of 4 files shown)


Arkham:
Test failed again due to github auth token missing, any way to fix that on travis?

lib/report.js
```diff
@@ -22,22 +22,46 @@ function jsonReport(errors) {
...
+                    path: errorForFile.path,
+                    ...error
+                  });
+                })
+                .join('\n');
             })
-          );
-        });
-      });
+            .join()
```
jfmengels, line 38:
```suggestion
            .join("\n")
```

This needs to be like this, otherwise the errors from different files will be joined by `,`

jfmengels (changes requested):
Thanks a lot for this!

> Test failed again due to github auth token missing, any way to fix that on travis?

That's weird, because the token is there :thinking: 
I'll look into switching to GitHub actions, since I'm moving everything there anyway for other repositories.

The tests pass for me locally anyway, so the CI is not a blocker.

Apart from that, there is one bug that was introduced but easily fixed.

jfmengels:
I've replaced Travis by GitHub actions on master, feel free to rebase (and I hope it will work)

lib/report.js
Arkham: Oh right, good catch!

Arkham:
Bug fixed and CI went green!

jfmengels (approved):
Awesome, thanks a lot @Arkham!

PR #28, opened 2020-11-21
Install elm-json via elm-tooling

lydell:
This PR uses [elm-tooling](https://github.com/lydell/elm-tooling.json/tree/main/cli) to install elm-json instead of the elm-json npm package.

Look at this size improvement:

| Version      |   Size |
| ------------ | -----: |
| 2.4.0-beta.2 |  20 MB |
| This PR      | 5.9 MB |

Size was measured by running `du -sh node_modules` after installing elm-review (and nothing else) on MacOS.

The elm-json binary itself around 2 MB on MacOS.

The huge space drop comes from not depending on `binwrap` anymore. It has some pretty have dependencies.

Benefits of using elm-tooling:

- Faster installs. Fewer npm dependencies to install. elm-review users often use elm-test as well – once both use elm-tooling there won’t be a double elm-json download on each `npm install`. And if the user uses elm-tooling to install elm and elm-format as well, all three tools are downloaded in parallel. [Similar PR for node-test-runner.](https://github.com/rtfeldman/node-test-runner/pull/480)
- Less disk space wasted: Less dependencies in node_modules, and no copy of elm-json in every project.
- Security. elm-tooling uses sha256 to verify that it downloads what it expects.

The only noticeable difference is _when_ the elm-json executable is transferred from the Internet to the user’s machine.

- Previously, it happened during `npm install`. Unless you use [ignore-scripts](https://docs.npmjs.com/using-npm/config#ignore-scripts) – then it happened the first time `elm-json` was executed. Now, it happens right before `elm-json` is about to be executed. Unless the user already downloaded elm-json to `~/.elm/elm-tooling/` before.

- Given that elm-review executes both elm and elm-json, which both can make network requests, I think it’s fine to also execute elm-tooling’s `getExectuable` function which also can make network requests. elm-review continues to work offline once all needed things exist on the user’s machine.

- binwrap uses the [request](https://github.com/request/request) module, while elm-tooling uses curl if available, otherwise wget and finally Node.js’ [https](https://nodejs.org/api/https.html) module. If the user used [request’s environment variables for proxies](https://github.com/request/request#controlling-proxy-behaviour-using-environment-variables), that should probably just work since [curl seems to use the same environment variables](https://ec.haxx.se/usingcurl/usingcurl-proxies#proxy-environment-variables). But there might be some little difference. Either way, both curl and wget also support config files that lets the user alter how they work as needed, which is a good thing.

- Previously, the elm-json binary was downloaded to `./node_modules/elm-json/unpacked-bin/`. Now it’s downloaded to `~/.elm/elm-tooling/elm-json/0.2.8/`. Elm and elm-json write to `~/.elm` too, so it should be fine. (elm-tooling supports `ELM_HOME` and Windows.)

The plan now is to:

- Try to transfer `elm-tooling` to the [@elm-tooling](https://github.com/elm-tooling/) org so I don’t become a single point of failure.
- Release elm-tooling 1.0.0. It works on Linux, macOS and Windows, has 100% test coverage, has been tested by a couple of people on Discord (Dillon Kearns even put it in his [elm-pages-starter](https://github.com/dillonkearns/elm-pages-starter) template) and I haven’t found any improvements to make in a month.
- Merge this PR and [the one for node-test-runner](https://github.com/rtfeldman/node-test-runner/pull/480).
- Announce `elm-tooling` in Slack and Discourse. So far I haven’t talked much about it, waiting for these two PRs to be made.


jfmengels:
I'll take a look at the PR once you give me the thumbs up.

I'm alreacy curious about a few things already:
- Do you know how to explain the size difference? Could the fact that it looks like you're now installing `elm` using `elm-tooling` too have an impact too? (I don't see `elm` as part of the dependencies anymore :thinking:)

- Can you think of some drawbacks/risks associated with switching to `elm-tooling`?
  - Any edge cases around running `elm-review`?
  - Any impact on a user's machine (installing `elm` to global locations for instance?)
  - Could some unexpected logs pop up when running `elm-review`?
  - If a user uses `elm-tooling` globally then uninstalls it or the binaries, would there be any surprises?

Thanks for the work, I'm willing to try it out but also a bit wary about any drawbacks that it may cause . But I guess we can always rollback if the experiment doesn't work out.

lydell:
> I'll take a look at the PR once you give me the thumbs up.

Great, I’ll let you know when it’s time! The plan is to make the equivalent PR to [node-test-runner](https://github.com/rtfeldman/node-test-runner) at the same time, release 1.0.0 of elm-tooling and see if we can transfer elm-tooling to the [@elm-tooling](https://github.com/elm-tooling/) org so I don’t become a single point of failure. My thinking is that it will be easier to do all of that once both PRs are up so I’ve got something concrete to show the involved parties.

> Do you know how to explain the size difference?

binwrap has a lot of heavy dependencies: https://packagephobia.com/result?p=binwrap

> Could the fact that it looks like you're now installing elm using elm-tooling too have an impact too?

elm was a devDependency before, so no.

> Any impact on a user's machine (installing elm to global locations for instance?)

elm-tooling writes to the same directory as elm and elm-json so it should be fine.

> Could some unexpected logs pop up when running elm-review?

elm-tooling’s `getExecutable` function writes nothing to stdout or stderr. We are in full control in the elm-review code.

> If a user uses elm-tooling globally then uninstalls it or the binaries, would there be any surprises?

No. The only thing that would happen would be that the `~/.elm/elm-tooling/` folder re-appears if the user previously removed it.

> Can you think of some drawbacks/risks associated with switching to elm-tooling?

The only noticeable difference is _when_ the elm-json executable is transferred from the Internet to the user’s machine.

- Previously, it happened during `npm install`. Unless you use [ignore-scripts](https://docs.npmjs.com/using-npm/config#ignore-scripts) – then it happened the first time `elm-json` was executed. Now, it happens right before `elm-json` is about to be executed. Unless the user already downloaded elm-json to `~/.elm/elm-tooling/` before.

- Given that elm-review executes both elm and elm-json, which both can make network requests, I think it’s fine to also execute `getExectuable` which also can make network requests. elm-review continues to work offline once all needed things exist on the user’s machine.

- binwrap uses the [request](https://github.com/request/request) module, while elm-tooling uses curl if available, otherwise wget and finally Node.js’ [https](https://nodejs.org/api/https.html) module. If the user used [request’s environment variables for proxies](https://github.com/request/request#controlling-proxy-behaviour-using-environment-variables), that should probably just work since [curl seems to use the same environment variables](https://ec.haxx.se/usingcurl/usingcurl-proxies#proxy-environment-variables). But there might be some little difference. Either way, both curl and wget also support config files that lets the user alter how they work as needed, which is a good thing.

jfmengels:
That all sounds very good :+1:

lydell:
I’ve updated the PR description to summarize everything, and opened a similar PR for elm-test: https://github.com/rtfeldman/node-test-runner/pull/480

Let’s discuss some more on Discord

.github/workflows/test.yml
```diff
@@ -18,10 +22,31 @@ jobs:
...
+      id: cache-node_modules
+      uses: actions/cache@v2
+      with:
+        path: node_modules
+        key: node_modules-ubuntu-${{ hashFiles('package-lock.json') }}
+
+    - name: npm ci
+      if: steps.cache-node_modules.outputs.cache-hit != 'true'
+      env:
+        NO_ELM_TOOLING_INSTALL: 1
```
jfmengels, line 45: What happens if `elm-tooling.json` is updated? For instance to say that we need `elm-format` v0.9.0. Shouldn't `elm-tooling` re-install the tools?

lib/template-dependencies.js
```diff
@@ -8,7 +8,7 @@ const crypto = require('crypto');
 const fs = require('fs-extra');
 const spawnAsync = require('cross-spawn-promise');
 const ErrorMessage = require('./error-message');
-const elmJsonCLI = require('elm-json').paths['elm-json'];
+const getExecutable = require('elm-tooling/getExecutable');
```
jfmengels, line 11: My personal advice would have been to expose `getExecutable` and all other utility functions in the `index.js` file of `elm-tooling`. If you put everything you want to expose as part of the index, then your API becomes much clearer.

If you allow importing from `elm-tooling/xyz`, then people will start doing that and the line between public and private API will become a lot more blurry.

lib/template-dependencies.js
```diff
@@ -111,11 +111,42 @@ async function update(options, pathToElmJson) {
...
+      name: 'elm-json',
+      version: '^0.2.8',
+      onProgress: (percentage) => {
+        const message = `Downloading elm-json... ${Math.round(
+          percentage * 100
+        )}%`;
+        process.stderr.write(
+          percentage >= 1
+            ? `${'Working...'.padEnd(message.length, ' ')}\r`
+            : `${message}\r`
```
jfmengels, line 128: Why stderr and not stdout? Also, does this update the message on the same line? I am having a hard time seeing how this would work out.

I think it would be better if we just returned a string and `elm-tooling` handles the writing to the standard output.

That might be a bad idea of mine, The API looks a bit implementation-sensitive to me. 

FYI, I personally like [`ora`](https://www.npmjs.com/package/ora), or the underlying [`cli-spinner`](https://www.npmjs.com/package/cli-spinner). Just showing them to you in case you didn't know and you might like it.

.github/workflows/test.yml
lydell: The tools listed in `elm-tooling.json` are installed in the next step. I’ve tried to explain CI setup here: https://github.com/lydell/elm-tooling.json/tree/main/cli#ci

lib/template-dependencies.js
lydell: `require("elm-tooling")` already means something else – it lets you easily run the `elm-tooling` CLI from a script: https://github.com/lydell/elm-tooling.json/tree/main/cli#elmtoolingcli

`require('elm-tooling/getExecutable')` is a bit “hidden” on purpose. Not many people will use that. In fact, I suspect nobody except elm-review and elm-test will use it. One thing I like with it being a separate entrypoint is that no unnecessary things are `require`:d.

I’ve got this in package.json:

```json
  "exports": {
    ".": "./index.js",
    "./getExecutable": "./getExecutable.js"
  },
```

So the public API is clearly defined, and newer Node.js versions won’t let people import anything else.

lib/template-dependencies.js
lydell: > Why stderr and not stdout?

`curl` prints its progress bar to stderr so I thought it was appropriate. I don’t care much about it – we can change it to stdout if you like.

> Also, does this update the message on the same line?

Yes. `\r` (Carriage Return) moves the cursor back to the start of the line.

> I think it would be better if we just returned a string and `elm-tooling` handles the writing to the standard output.

Why? A function that gets the percentage as a number and allows doing anything you want is the simplest and most flexible API I can think of. Why restrict people to only being able to print stuff to `stdout`? What if the want to print to `stderr` or show progress some completely different way?

> FYI, I personally like ora, or the underlying cli-spinner. Just showing them to you in case you didn't know and you might like it.

I just did the simplest possible progress output to get a discussion going. I was hoping you’d comment on how you prefer it to be shown. Let me know how you want it and I’ll change. Also feel free to push directly to this branch if you feel like fiddling with it yourself.

(Maybe for inspiration, in node-test-runner I added progress output not just for downloading elm-json, but for the whole run. For example, `Solving dependencies > Compiling > Starting tests` where each segment appears over time.)

lib/template-dependencies.js
jfmengels: Oh man, I didn't follow this improvement in Node.js, great! (I should stop burying my head in Elm-land)

.github/workflows/test.yml
jfmengels: Ah, so setting `NO_ELM_TOOLING_INSTALL` doesn't prevent the next step.
Should the result of `elm-tooling install` not be cached?

lydell: > Ah, so setting NO_ELM_TOOLING_INSTALL doesn't prevent the next step.

Correct, were setting env vars per step here, and they don’t propagate to the next steps.

> Should the result of `elm-tooling install` not be cached?

`elm-tooling install` downloads stuff to `~/.elm/elm-tooling/` and we cache `~/.elm` above.

`elm-tooling install` also generates “links” to the downloaded binaries in `./node_modules/.bin/`, but that’s so fast that trying to avoid that work through caching doesn’t bring any time savings.

lib/template-dependencies.js
jfmengels: Yeah. Let's leave it for now, and I'll change it to whatever feels nicest later. I think you're right, that this is the best API for it, since we're not even necessarily in the context of a terminal.

I will change it so it doesn't print anything when running with `--report=json` though.

Thanks for the info on `\r`, good to know!

.github/workflows/test.yml
jfmengels: > elm-tooling install downloads stuff to ~/.elm/elm-tooling/ and we cache ~/.elm above

Alright, I thought that the action cached the contents of `~/.elm` as it was at the moment that step was run. So since `elm-tooling` downloads the binaries later on, I thought we cached an empty folder.

But it caches it at the end of the run but restores the cache at this step, is that right?

jfmengels:
I think this looks good. What do you think we should do with the `elm` and `elm-format` binaries? Do you think we should install/use them through `elm-tooling` too, but maybe at a later date?

This is mergeable as is in your opinion, right?

.github/workflows/test.yml
lydell: That is correct! You can see that it works that way here: https://github.com/jfmengels/node-elm-review/pull/28/checks?check_run_id=1436496301

lydell:
> What do you think we should do with the elm and elm-format binaries? Do you think we should install/use them through elm-tooling too

This pull request already switched to installing elm and elm-format via elm-tooling.json (for development of node-elm-review itself).

> This is mergeable as is in your opinion, right?

Yes, it’s mergeable! You can either wait for elm-tooling 1.0.0 to be released, or merge now (if so I’ll make a new PR updating to elm-tooling 1.0.0 when it’s out).

jfmengels:
> This pull request already switched to installing elm and elm-format via elm-tooling.json (for development of node-elm-review itself).

I meant for non-development purposes:
- Elm: compiling the configuration (using the version we find in the user's `review/elm.json` file)
- elm-format: Run elm-format based on the version found in `elm-tooling.json` (and default to whatever `elm-tooling` would choose in the absence of that file).

both of these would be overriden by the flags passed to `elm-review` to specify the paths to these tools. What do you think? Anyway, that's can be done later, as that is a very different use-case than the intention to replace `elm-json`.

I feel like I've been annoying you with this PR, sorry for all that :sweat_smile: , and thanks a lot with your work, which I found to be of very high-quality :blush: (especially since you were able to answer all my worries flawlessly :D )

lydell:
> I feel like I've been annoying you with this PR, sorry for all that 😅

No, not at all! I’m sorry for having sounded annoyed. Sometimes short Github comments is such as crude form of communication. I think it’s very good you review stuff like this very carefully, and I didn’t mind answering your questions at all.

About running `elm` and `elm-format` based on the versions found in `elm-tooling.json` – we already indirectly support that. When users run `elm-tooling install`, `./node_modules/.bin/elm` and `./node_modules/.bin/elm-format` are created (they are symlinks to the actual binaries – just like npm puts symlinks to stuff in `./node_modules/.bin/`). But maybe it would be better to support it in a more direct way – look for `elm-tooling.json`, and if it exists parse it and look for versions and resolve the paths etc… I haven’t given this enough thought yet.

lydell:
I’ve been thinking about if elm-review should be aware of elm-tooling.json when it invokes `elm` and `elm-format`. It’s also interesting to swap elm-review for **elm-test-rs** when thinking about this. Sorry for the rambly wall of text.

I’m not sure if tools _should_ explicitly care about elm-tooling.json.

I _think_ tools should just assume that `elm` (or `elm-format`) exists in `$PATH` so they can just try to `spawn('elm')` or whatever.

Why?

- If you install `elm-review` locally, you’ll be running it using `npx elm-review` or `npm run elm-review` (with a script in package.json) or `yarn elm-review` or whatever. In all those cases, npx/npm/yarn adds `./node_modules/.bin` to `$PATH` for the command they spawn. So according to `elm-review`, `elm` (and `elm-format`) do exist in `$PATH` when it runs, no matter if `elm` (and `elm-format`) were installed locally or globally.

- If you install `elm-test-rs` locally using `elm-tooling` (once I’ve added support for it), the same thing will happen. You’ll run it using `npx elm-test-rs` (or one of the other alternatives), again causing stuff from `./node_modules/.bin` be available in `$PATH` for that command.

This is why I went with the approach I did in `elm-tooling`: I both download executables to `~/.elm/elm-tooling`, and create links to them in `./node_modules/.bin`. This was mostly because it makes `elm-tooling` a drop-in replacement for `npm install elm elm-format`. But now I’m starting to think that this “local linking” might be a good approach in a larger picture.

There’s something pleasingly simple in tools being able to just do stuff like `spawn('elm')` and don’t have to worry about where the `elm` executable is located or where it came from. (In practice, it would be `spawn(options.compiler)`, because I think it’s good practice to have a `--compiler /path/to/elm` CLI flag for special cases.)

Maybe that’s better than trying to use `elm-tooling.json` more “directly”? For comparison, a more “direct” use would involve:

1. Finding an `elm-tooling.json` up the folder hierarchy.
2. Try to parse it. What should one do if there’s an error in there?
3. Look at the `"tools"` field and resolve the path to `elm` (or whatever tool you’re looking for) according to the version specified in there. This involves reading the `ELM_HOME` environment variable, and defaulting to either `~/.elm` or `%APPDATA%\elm`.
4. See if the executable exists. What if it doesn’t?
5. Finally, run the executable.
6. If no elm-tooling.json was found, fall back to other ways of running `elm`. I suppose you’d want to support `./node_modules/.bin/elm` and a global `elm`.

`elm-tooling` could ship with an API function for doing all/most of that, but there’s some questions around that.

- How much should it do, and how much should be left to the consumer?
- What about elm-test-rs which isn’t written in JS (it’s written in Rust)? Or a tool written in Haskell?

Final thought. What if `elm-tooling` was written in, say, Rust, instead of in JS (Node.js). Then it would have no relation to `npm` anymore. How would people run downloaded tools then? Would there still be links in `./node_modules/.bin/`? Would people still use `npx elm`? But what if they don’t even have Node.js/npm installed and don’t want it?

Maybe there would be an `elm-tooling run elm` command. It would do all the lookups mentioned above. And run the executable with a `$PATH` containing all the local tools. So `elm-tooling run elm-test-rs` would execute the local version of `elm-test-rs` with a `$PATH` containing the local versions of `elm` and `elm-format`.

But what about IDEs/editors? They already support detecting `./node_modules/.bin/elm` etc, so maybe they could support using `elm-tooling` to execute `elm` etc?

In conclusion, I have a _ton_ of thoughts floating around, but it _feels_ like it could be a reasonable and simple solution to _just leave everything as it is._ Tools should just assume that they have `elm` etc in `$PATH`. And there seems to be a future for that approach even if `npm` at some point vanishes from the picture.

PR #30, opened 2021-01-09
Update elm-tooling to 1.0.0

lydell:


jfmengels:
Oh I hadn't seen you made this pull request!

I just pushed a change to bump it to 1.0.1, so I'll be using that instead.
Thanks though! :)

PR #31, opened 2021-01-22
Improve elm-json stderr message extraction

lydell:
Previously, if elm-json failed to connect to package.elm-lang.org (in an environment with no cache) the following error message was printed:

```

  jfmengels/node-elm-review/issues/new

Please include this error message and as much detail as you can provide. If you
can, please provide a setup that makes it easy to reproduce the error. That will
make it much easier to fix the issue.

Error: Exited with status 1
    at closeArgsToError (/nix/store/hw5vcijxs0d83gwn8f220bqif3f47kl5-node_elm-review-2.4.0/lib/node_modules/elm-review/node_modules/cross-spawn-promise/lib/index.js:20:16)
    at ChildProcess.<anonymous> (/nix/store/hw5vcijxs0d83gwn8f220bqif3f47kl5-node_elm-review-2.4.0/lib/node_modules/elm-review/node_modules/cross-spawn-promise/lib/index.js:76:19)
    at Object.onceWrapper (events.js:422:26)
    at ChildProcess.emit (events.js:315:20)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12)
```

Now it looks like this instead:

```
-- CONFIGURATION COMPILATION ERROR ---------------------------------------------

I encountered a problem when solving dependencies:

-- NO VALID PACKAGE VERSION ----------------------------------------------------

Because elm/json does not appear to exist and this project depends on elm/json
at any version, no valid set of package versions could be found.

phase: retrieve
 Jan 22 18:15:41.748 WARN Failed to fetch versions from package.elm-lang.org
```

While at it, I replaced `cross-spawn-promise` with just `cross-spawn` (which was already a dependency). `cross-spawn-promise` depends on `cross-spawn` 5.x, while the latest `cross-spawn` is 7.x.

### DIFF
--- a/lib/template-dependencies.js
+++ b/lib/template-dependencies.js
@@ -6,7 +6,7 @@ const path = require('path');
 const util = require('util');
 const crypto = require('crypto');
 const fs = require('fs-extra');
-const spawnAsync = require('cross-spawn-promise');
+const spawnAsync = require('cross-spawn');
 const ErrorMessage = require('./error-message');
 const getExecutable = require('elm-tooling/getExecutable');
 
@@ -71,14 +71,16 @@ ${formatElmJsonError(error, options)}`,
 }
 
 function formatElmJsonError(error, options) {
-  if (error.stderr === undefined || error.stderr === null) {
+  if (error.stderr === undefined) {
     return error.message;
   }
 
   const stderrMessage = error.stderr.toString().trim();
   const exec = /^([^]+\n)?--\s([A-Z ]+)\s-*\n\n([^]+)$/.exec(stderrMessage);
   if (exec === null) {
-    return `${error.message}\n\n${stderrMessage}`;
+    return `${error.message}\n\n${
+      stderrMessage === '' ? '(empty stderr)' : stderrMessage
+    }`;
   }
 
   const [, before, title, message] = exec;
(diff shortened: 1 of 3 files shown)


lib/template-dependencies.js
```diff
@@ -82,6 +70,29 @@ ${formattedError}`,
     .catch(handleInternetAccessError);
 }
 
+function formatElmJsonError(error, options) {
+  if (error.stderr === undefined) {
+    return error.message;
+  }
```
lydell, line 76: You never know what you catch in JS. Just in case this error didn’t come from our handler which attaches `.stderr`, return the error message rather than crashing (“cannot call .toString() on undefined”.)

lib/template-dependencies.js
```diff
@@ -82,6 +70,29 @@ ${formattedError}`,
...
+
+  const stderrMessage = error.stderr.toString().trim();
+  const exec = /^([^]+\n)?--\s([A-Z ]+)\s-*\n\n([^]+)$/.exec(stderrMessage);
```
lydell, line 79: I tweaked the regex to support stuff being printed before the `-- HEADER -----` as well, as in:

```
phase: retrieve
 Jan 22 18:15:41.748 WARN Failed to fetch versions from package.elm-lang.org

-- NO VALID PACKAGE VERSION ----------------------------------------------------

Because elm/json does not appear to exist and this project depends on elm/json
at any version, no valid set of package versions could be found.
```

jfmengels: Regex question: I don't understand what the `[^]` at the beginning does, and https://regex101.com/ seems to have trouble parsing this regex.

lydell: It’s one of the many ways to match _anything._ Your `(.|\n)` approach missed some characters (such as `\r`). Writing this I realize the `/s` (dotAll) flag has good support these days and we could have used that instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#browser_compatibility

PR #32, opened 2021-01-29
Consider when elm-tooling is running in "CI mode"

gpampara:
In the situation where `elm-tooling` in running in [ci mode](https://elm-tooling.github.io/elm-tooling-cli/ci/), a user can set an environment variable and `elm-tooling` will assume that the dependencies are implicitly available on the system. In this case, either the system tools are used or the run will fail.

This PR let's `elm-review` consider the CI usage of `elm-tooling` by checking that the return result from the call to `getExecutable` is the integer 0, as opposed to the expected path string.

I'd be more than happy to rework this based on any feedback.

### DIFF
--- a/lib/template-dependencies.js
+++ b/lib/template-dependencies.js
@@ -176,39 +176,42 @@ ${error.message}`
     });
   }
 
-  return elmJsonPromise.then(
-    (elmJsonCLI) =>
-      new Promise((resolve, reject) => {
-        const child = spawnAsync(elmJsonCLI, args, {
-          silent: true,
-          env: process.env
-        });
-        let stdout = '';
-        let stderr = '';
+  return elmJsonPromise.then((elmJsonCLI) => {
+    // For CI elm-tooling returns the value of 0. CI environments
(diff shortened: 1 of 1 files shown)


lib/template-dependencies.js
```diff
@@ -176,39 +176,42 @@ ${error.message}`
...
-            error.stderr = stderr;
-            reject(error);
-          }
-        });
-      })
-  );
+  return elmJsonPromise.then((elmJsonCLI) => {
+    // For CI elm-tooling returns the value of 0. CI environments
+    // are defined when the env var NO_ELM_TOOLING_INSTALL exists
+    if (elmJsonCLI === 0) return;
```
lydell, line 182: `elmJsonCLI` is _always_ a string here (an absolute path to the `elm-json` executable), or am I confused?

gpampara: @lydell I think that this is an error on my side - there is something weird with this bunch of node_modules. I'm going to dig into this a bit more and will reopen this PR if needed,

PR #43, opened 2021-04-03
Wrap elm-format filepath argument so the paths can include whitespace

thomasin:
Hello (: I store some of my projects in a path that includes a whitespace (🏄‍♀️). When running `elm-review --fix-all`, it passes each file to `elm-format` but doesn't quote them, which causes the command to fail with a "Can't find file" error. This change wraps the file path argument in quotes. I am also happy to add tests, but wanted to see what you thought first.
Cheers


jfmengels:
This looks very good and I can confirm locally that it fixes the problem :)
I will add tests myself. Thanks for the PR! :heart: 

I will make a release likely in the next few days, I hope you can wait until then :pray:

PR #48, opened 2021-05-28
Improve error message when entering an invalid rule name

### LINKED ISSUE #47: `elm-review new-package` does not allow module names with non-ASCII in them
```
✔ Name of the rule (ex: No.Doing.Foo): … Åäö
The rule name needs to a valid module name.
```

But `Åäö` _is_ a valid module name!

In practice, I would never use non-ASCII for module names, but the Elm compiler allows it.

I think elm-review should either allow any valid module name, or change the error message.

I you want regex inspiration, look here (that’s not a regex for module names with dots, but for a single name starting with an uppercase letter):

https://github.com/rtfeldman/node-test-runner/blob/11733ef616846973898770326cc0682f07eb984a/lib/Parser.js#L234

jfmengels:
Fixes #47 

Instead of allowing unicode characters, I improved the error message saying what isn't allowed. We noticed that some tools, like Git, don't support "exotic" characters too well (in things like `git status -s`, and allowing these characters would make it harder for other people to use a published role.

People can still workaround this by renaming the module if they really want to.

### DIFF
--- a/lib/new-rule.js
+++ b/lib/new-rule.js
@@ -42,7 +42,7 @@ function readElmJson(options) {
   }
 }
 
-const ruleNameRegex = /^[A-Z][\d\w_]*(\.[A-Z][\d\w_]*)*$/;
+const ruleNameRegex = /^[A-Z]\w*(\.[A-Z]\w*)*$/;
 
 function validateRuleName(ruleName) {
   if (!ruleNameRegex.test(ruleName)) {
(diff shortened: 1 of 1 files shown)


lydell:
Nice!

Not related to the changes in this PR, but as an FYI:

```js
const ruleNameRegex = /^[A-Z][\d\w_]*(\.[A-Z][\d\w_]*)*$/;
```

`\w` means `[A-Za-z0-9_]` so `[\d\w_]` can be simplified to just `\w`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Character_Classes

jfmengels:
Good catch. Looks like I could use a linter for that :roll_eyes:

lydell:
There’s https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/better-regex.md

I haven’t tried it myself, though

PR #50, opened 2021-06-26
Error in "successive releases" example script?

mthadley:
I had trouble running the `npm run elm bump` line in the script for successive releases. Is there a missing `elm-bump` package script? Or perhaps some `elm-bump` tool (à la `elm-json`) that I'm not aware of?

Either way, I found that `npx elm bump` and `yarn elm bump` both do the trick here, so I figured I'd at least propose it.

P.S. Thanks for `elm-review. This package (and tools you bundle to maintain rules) have been :exploding_head: my mind!

### DIFF
--- a/new-package/maintenance/MAINTENANCE.md
+++ b/new-package/maintenance/MAINTENANCE.md
@@ -117,7 +117,7 @@ Contrary to the initial release, the CI will automatically try to publish a new 
 Here is a script that you can run to publish your package, which will help you avoid errors showing up at the CI stage.
 
 ```bash
-npm run elm-bump
+npx elm bump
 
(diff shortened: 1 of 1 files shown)


mthadley:
Nevermind! Looks like I wasn using an older version of the `package.json` template (which I had copied from elsewhere after I accidentally deleted it). [There's clearly a `elm-bump` script in the template](https://github.com/jfmengels/node-elm-review/blob/c85e9252e319c80583164a4c343ca62dd03b9473/lib/new-package.js#L351). I'll close this.

jfmengels:
Glad you got it figured out, and that you like the project :blush:

PR #53, opened 2021-07-31
don't assume path `*.elm` is a regular file

### LINKED ISSUE #52: Does not support directory ending in '.elm'
We have a convention of putting elm files in a folder called `src.elm` (to go alongside `src.js` for port code). But this doesn't appear to work with elm-review.

```
$ elm-review
-- NO FILES FOUND --------------------------------------------------------------

I was expecting to find Elm files in all the paths that you passed, but I could
not find any in the following directories:
- /home/db/app/src.elm

When I can’t find files in some of the directories, I’m assuming that you
misconfigured the CLI’s arguments.
```

I believe it's due to this special case: https://github.com/jfmengels/node-elm-review/blob/b3ae61b0e600fde3887ae344e2e0e2305ec73928/lib/elm-files.js#L215-L216

Maybe `if (path_.endsWith('.elm'))` should be replaced by checking if the path points to a regular file?

dbandstra:
Fixes #52

### DIFF
--- a/lib/elm-files.js
+++ b/lib/elm-files.js
@@ -212,7 +212,8 @@ async function findFiles(directory) {
   const path_ = directory.replace(/.:/, '').replace(/\\/g, '/');
   const ignore = ['**/elm-stuff/**'];
 
-  if (path_.endsWith('.elm')) {
+  const stat = await fsStat(directory);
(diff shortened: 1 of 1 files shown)


dbandstra:
Superseded by #54

PR #54, opened 2021-08-02
Check whether a source directory is a file by checking it with the file system

### LINKED ISSUE #52: Does not support directory ending in '.elm'

jfmengels:
Fixes #52

@dbandstra I added some tests and noticed your solution didn't work out for several reasons (though I did what you suggested, so it was useful, thanks :) )
- The `fsStat` was indeed running even without passing arguments
- Therefore when you don't add arguments, `tests/` gets added automatically and if you don't have any, there is a crash about it not existing.

If you could test this out that'd be great :blush:

### DIFF
--- a/lib/elm-files.js
+++ b/lib/elm-files.js
@@ -251,8 +251,12 @@ async function findForCliArguments(directory, path_) {
     return findFromFolder(directory, path_);
   } catch (error) {
     if (error.code === 'ENOENT') {
-      // File was not found
-      return undefined;
+      // File/folder was not found. Return an empty array, which will give a nice
+      // error report in the parent function
+      return {
+        files: [],
+        directory
+      };
     }
 
     throw error;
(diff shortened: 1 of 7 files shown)


dbandstra:
It didn't occur to me that the "maybe a file" logic shouldn't apply to source-directories... seems obvious in retrospect. And good catch with the `tests` not existing, I totally missed that.

I tested your branch out with my project and it works great. Thanks!

jfmengels:
Alright, thanks for checking :pray:

PR #55, opened 2021-08-22
Update and fix elm-tooling usage

lydell:
elm-tooling 1.4.0 accidentally shipped with faulty exports, which was fixed in 1.4.1: https://github.com/elm-tooling/elm-tooling-cli/blob/b292a8169019e5ef5b348a0d006807ca8472c5d8/CHANGELOG.md#version-141-2021-08-22

This PR makes sure 1.4.1 or newer is used, and reverts https://github.com/jfmengels/node-elm-review/commit/5f81e8ebdd05894fc8ed349d93bce9b83e96dd6e (which accounted for the faulty export).

### DIFF
--- a/lib/new-package.js
+++ b/lib/new-package.js
@@ -3,7 +3,7 @@ const childProcess = require('child_process');
 const chalk = require('chalk');
 const fs = require('fs-extra');
 const prompts = require('prompts');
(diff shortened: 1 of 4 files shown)


jfmengels:
Thanks a lot 🙏

PR #56, opened 2021-08-22
Fix failing `ReviewConfig` for `new-package`

SiriusStarr:
8dd7613 added `Simplify.defaults` but `elm.json` still had "jfmengels/elm-review-simplify": "1.0.2", while `Simplify.defaults` was added in `2.0.0`

This bumps to the latest `elm-review-simplify` version to fix.


jfmengels:
Aha! Thanks a lot!

PR #59, opened 2021-09-28
new-package: Add `elm-format` checks for preview files

SiriusStarr:
Currently, the default tests for packages do not check if the Elm files in e.g. `preview/` are formatted with `elm-format` (though they do check if they compile).  This adds just such a check.

Note that `preview*/**.elm` is used rather than `preview*/`, because otherwise an empty folder will cause it to error-out.  This solution seems to work well for relatively complex preview structures.

### DIFF
--- a/lib/new-package.js
+++ b/lib/new-package.js
@@ -343,7 +343,7 @@ function packageJson(options, packageName) {
       test:
         'npm-run-all --print-name --silent --sequential test:make test:format test:run test:review test:package',
       'test:make': 'elm make --docs=docs.json',
-      'test:format': 'elm-format src/ tests/ --validate',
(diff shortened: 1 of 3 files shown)


jfmengels:
It seems (haven't tried but did ask people) that `**.elm` does not work on Windows, so I'd prefer going for `preview*/`. Have you run in practice into issues with the empty folder?

SiriusStarr:
No, I can't imagine that it would be an issue in practice, I was just trying not to impose an additional (unintended) constraint that didn't already exist.  If you're good with requiring that there be no folders caught by the `preview*`glob that lack Elm files, that's probably the best option.  TBH I forget Windows exists as a dev environment; didn't know it can't handle `**.elm`.  :rofl:

jfmengels:
This looks good, thank you! :blush:

PR #64, opened 2021-11-21
Change output colours to support light background

alexkorban:
Not a lot of colours to choose from; I propose the following changes as they work on both light and dark backgrounds and don't seem to affect the existing semantics of colours (too much).

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -90,9 +90,9 @@ async function buildLocalProject(options) {
         new ErrorMessage.CustomError(
           /* eslint-disable prettier/prettier */
 'INCORRECT CONFIGURATION',
-`I could not find a review configuration. I was expecting to find an ${chalk.yellowBright('elm.json')} file and a ${chalk.cyan('ReviewConfig.elm')} file in ${chalk.cyan(path.dirname(reviewElmJsonPath))}${chalk.cyan('/')}.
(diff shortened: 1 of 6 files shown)


jfmengels:
> Not a lot of colours to choose from

I know :sweat_smile: 

I think I'm generally ok with the change. Thanks a lot for looking at this :pray: 

I noticed that in this case you change the cyan to a magenta (purple), but you kept the yellow. Is that still readable for you? I'd say it is, but I prefer checking with you.

Screenshot from 2021-11-22 17-18-50

When trying it out, that gives me this (which gives me the feeling that any color on a white background, while readable, is not great :sweat_smile: )

Screenshot from 2021-11-22 17-21-43

If you deem this yellow to be readable, then maybe we could use that color where you've removed it before (like in the `--help`).


Also, there are other usages [of yellow in the Elm part](https://github.com/jfmengels/node-elm-review/blob/master/template/src/Elm/Review/Text.elm#L99-L101), which can be seen at the end of this output.

Screenshot from 2021-11-22 17-25-49

(By the way, to try this out, I recommend running `<path-to-local-elm-review` --force-build` when you touch the Elm code in `node-elm-review`, otherwise it will use a cached version).

alexkorban:
I thought `chalk.yellow` is readable enough so I only wanted to change `chalk.yellowBright`. I've now changed `yellowBright` to `yellow` instead of `green` - happy with that if you think it works semantically. I'm not sure there's a solution that pleases everyone universally because people might have set any background colour, not just black or white.

Also updated the yellow in `Text.elm`; I picked a similar looking colour visually as I wasn't quite sure what the exact value of `chalk.yellow` is. Hope that works.

jfmengels:
I think these additions are good. Let's try them out :)
Thanks for the PR! (and sorry for the delay).

PR #100, opened 2023-01-25
Fix `check-previews-compile.js` failure on long JSON output

SiriusStarr:
Per discussion on the slack, this fixes the issue of JSON output truncating at a certain length.  Why is beyond me and known only to the javascript gods.  Credit to @sparksp for figuring it out.

### DIFF
--- a/new-package/elm-review-package-tests/check-previews-compile.js
+++ b/new-package/elm-review-package-tests/check-previews-compile.js
@@ -21,7 +21,7 @@ function checkThatExampleCompiles(exampleConfiguration) {
   );
 
   try {
-    execSync(`npx elm-review --config ${exampleConfiguration} --report=json`, {
+    execSync('npx elm-review', ['--config', exampleConfiguration, '--report=json'], {
       encoding: 'utf8',
       stdio: 'pipe',
       cwd: path.resolve(__dirname, '..')


new-package/elm-review-package-tests/check-previews-compile.js
```diff
@@ -21,7 +21,7 @@ function checkThatExampleCompiles(exampleConfiguration) {
   );
 
   try {
-    execSync(`npx elm-review --config ${exampleConfiguration} --report=json`, {
+    execSync('npx elm-review', ['--config', exampleConfiguration, '--report=json'], {
```
lydell, line 24: `execSync` does not take an array of options.

https://nodejs.org/api/child_process.html#child_processexecsynccommand-options

Here you can see how passing 3 arguments results in the third being ignored, and passing an array as the second argument is basically the same as passing `{}` (empty options) there.

```
node
> child_process.execSync('echo a b c')
<Buffer 61 20 62 20 63 0a>
> child_process.execSync('echo a b c', {encoding: 'utf8'})
'a b c\n'
> child_process.execSync('echo a', ['b', 'c'], {encoding: 'utf8'})
<Buffer 61 0a>
> child_process.execSync('echo a', ['b', 'c'], {encoding: 'utf8'}).toString()
'a\n'
```

lydell: I haven’t read through the code enough here but there are many alarm bells ringing with the original code here:

1.  String interpolation in a shell command. Is spaces in the interpolated value going to mess it up?
2. The use of `npx`.
3. The use of `child_process` directly – does it work on Windows?

sparksp: Good catch! Looks like it's a false positive fix then. Sorry all!

SiriusStarr:
Nevermind, this doesn't fix anything; someone who actually understands javascript will have to figure out a fix to #101

Issue #105, opened 2023-02-28
check-previews-compile.js requires glob < 9


sparksp:
New major version of [glob v9.0.1](https://www.npmjs.com/package/glob/v/9.0.1) has changed the interface which will break `elm-review-package-tests/helpers/find-configurations.js`. Hopefully a fairly easy fix.

```
./elm-review-always/elm-review-package-tests/helpers/find-configurations.js:15
    .sync(`${root}/preview*/**/elm.json`, {
     ^

TypeError: glob.sync is not a function
    at findPreviewConfigurations (./elm-review-always/elm-review-package-tests/helpers/find-configurations.js:15:6)
    at Object.<anonymous> (./elm-review-package-tests/check-previews-compile.js:12:1)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Module.load (node:internal/modules/cjs/loader:1117:32)
    at Module._load (node:internal/modules/cjs/loader:958:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47
```

jfmengels:
Fixed in 58e1ebfb614bb233a1f8158f36c1a9766a172dbd

PR #106, opened 2023-02-28
Update glob to v9

sparksp:
The outstanding issue here is that package authors will need to update `new-package/elm-review-package-tests/helpers/find-configurations.js` themselves when they update glob to v9, as this was where I had issues. I'm not sure how best to communicate this though? Can/should `elm-review` provide an `update-package-tests` command? Or is it something for the README or other documentation?

Fixes https://github.com/jfmengels/node-elm-review/issues/105


sparksp:
[glob@9.3.0](https://github.com/isaacs/node-glob/blob/main/changelog.md#93) adds aliases including `glob.sync` so it's no longer necessary to make these code changes, just update glob to at least 9.3.0.

jfmengels:
Done in 58e1ebfb614bb233a1f8158f36c1a9766a172dbd, thanks for looking into this :pray:

PR #112, opened 2023-04-15
Separate fetching rule links and project dependencies

jfmengels:
These things are pretty different, and don't really need to be shared.

### Context

Project dependencies are used for the analysis of the project

Rule links are information fetched to create links on the rule name in the terminal that send you to the rule's documentation, like in the screenshot below:

image

### Effects

1) Separating the fetching of these will make it easier in future work to only fetch the updated project dependencies when they have changed, without fetching the rule links.

2) The code should be slightly easier to read, as I've split things up into multiple self-contained modules.

3) Startup performance might be a bit improved because of a few changes:
- Since these two steps are now separated, we can start the fetching of rule links earlier, before some necessary steps for fetching the project deps are completed
- We now re-use the same promise for fetching the same dependencies (should be rare though!)
- We don't fetch the `elm/*` dependencies for rule links, as it's quite unlikely that they will contain `elm-review` rules in the foreseeable future.

### DIFF
--- a/lib/project-dependencies.js
+++ b/lib/project-dependencies.js
@@ -1,13 +1,4 @@
-const path = require('path');
-const os = require('os');
 const ProjectJsonFiles = require('./project-json-files');
-
-const elmRoot =
-  process.env.ELM_HOME ||
-  path.join(
-    os.homedir(),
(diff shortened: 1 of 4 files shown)


jfmengels:
Merged into master.

PR #113, opened 2023-04-15
Re-organize cache logic structure

jfmengels:
- Extract the duplicated hash function to a new Hash module
- Rename the Cache module to ModuleCache
- Create some helpers to cache JSON files in a "new" Cache module
- 

### Misc

- Rename some imports to be capitalized


jfmengels:
Merged into master.

PR #114, opened 2023-04-15
Collect dependencies when elm.json gets autofixed

jfmengels:
When `elm.json` gets autofixed because of rules like `NoUnused.Dependencies`, the list of dependencies in the project will be refetched.

This should remove potential false positives where a rule takes the list of dependencies as a source of truth instead of the dependencies listed in `elm.json`.

Some further work can be done later to avoid doing this when the dependencies fields have not changed. And similar work should be done for when the source-directories change. But so far the only known rule that autofixes `elm.json` is `NoUnused.Dependencies`, which does not trigger false positives even if the dependencies were not getting updated.


jfmengels:
Merged into master.

PR #122, opened 2023-04-22
Add typescript

jfmengels:
The purpose of this PR is to use TypeScript to type-check the JS code of the CLI, to feel more at ease when changing that part of the codebase, which todays feels quite scary to be honest.

It's apparently quite do-able to have TS understand type annotations [through JSDoc comments](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html), so I think I want to try that instead of converting files to `.ts`. If that works out well, then we can skip the transpilation phase, which will keep it pretty smooth to run the CLI in development mode (no need to wait for TS to compile the project).

We can potentially also use this to publish `.d.ts` file, in case someone wants to depend on the JS API of the tool (like https://github.com/elm-tooling/elm-language-server/pull/942). I imagine this would help, but I could be wrong.

## Status of the PR

I'm finding a bunch of issues that I add to the main branch, but this PR is not ready yet.
First of all, I have a large number of problems that TS reports that seem like configuration errors.
Second, I still need to add TS to the test script so that we're made aware of TS errors.

In a separate PR, we could then remove Flow.

I'd appreciate any help with this work and/or with adding type annotations to the JS codebase (better done in separate commits).


tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "target": "es2015",
+    "moduleResolution": "node",
+    "noEmit": true
```
lydell, line 9: Just in case you didn’t know: At some point you probably want to make the typechecking more comprehensive, especially adding `"strict": true`. More candidates:

```json
    "strict": true,
    "noFallthroughCasesInSwitch": true,
    "noImplicitReturns": true,
    "useUnknownInCatchVariables": true
```

jfmengels: Thanks! I knew I didn't check for much (because I first want to get rid of the configuration-related errors), but I didn't know exactly which things I needed to be enable. This saves me some times :blush: 
If you have more tips and candidates, I'll take them. I think there was one about implicit any for instance, which I'll try to find at some point.

lydell: `strict` is a shortcut for enabling a bunch of other options, including `noImplicityAny`. I couldn’t find the exact list of options in TypeScript’s documentation, but I found this in a blog post:

```
    noImplicitAny
    noImplicitThis
    alwaysStrict
    strictBindCallApply
    strictNullChecks
    strictFunctionTypes
    strictPropertyInitialization
```

I went through the [elm-watch tsconfig.json](https://github.com/lydell/elm-watch/blob/main/tsconfig.json), and the last strictness related one seems to be:

```json
    "noUncheckedIndexedAccess": true,
```

But when adding TypeScript to an already existing code base it might make sense to enable one thing at a time so you don’t drown in errors, before you have even learned how to deal with many of them.

The final level is to bring in typscript-eslint. Then you get close to Elm level type system confidence.

jfmengels: Perfect, thanks :)

jfmengels:
@lydell the only remaining errors I have (well, before I enforce new things), are the following ones related to dependencies. I think TS is wrong about these (I mean, they work so...). Would you know how I can correct its findings?

```
lib/new-package.js:174:9 - error TS2349: This expression is not callable.
  Type 'typeof import("/home/jeroen/dev/node-elm-review/node_modules/elm-tooling/index")' has no call signatures.

174   await elmToolingCli(['init'], {
            ~~~~~~~~~~~~~

lib/project-json-files.js:119:26 - error TS2349: This expression is not callable.
  Type 'typeof import("/home/jeroen/dev/node-elm-review/node_modules/got/dist/source/index")' has no call signatures.

119   const response = await got(
                             ~~~

lib/remote-template.js:106:10 - error TS2349: This expression is not callable.
  Type 'typeof import("/home/jeroen/dev/node-elm-review/node_modules/got/dist/source/index")' has no call signatures.

106   return got(
             ~~~

lib/remote-template.js:260:10 - error TS2349: This expression is not callable.
  Type 'typeof import("/home/jeroen/dev/node-elm-review/node_modules/got/dist/source/index")' has no call signatures.

260   return got(url, parameters)
             ~~~

lib/template-dependencies.js:202:22 - error TS2349: This expression is not callable.
  Type 'typeof import("/home/jeroen/dev/node-elm-review/node_modules/elm-tooling/getExecutable")' has no call signatures.

202     elmJsonPromise = getExecutable({
                         ~~~~~~~~~~~~~
```

jfmengels:
I found solutions to some of these last issues, and ts-ignored the rest :(

jfmengels:
I found solutions to some of these last issues, and ts-ignored the rest :(

lydell:
@jfmengels Try adding `"esModuleInterop": true,` to tsconfig.json. I think that should do it for both elm-tooling and got.

lydell:
Nope, I was wrong about `"esModuleInterop": true`. It has helped me before with dependencies in other projects though!

For me, this works:

```js
const { default: elmToolingCli } = require("elm-tooling");
const { default: getExecutable } = require("elm-tooling/getExecutable");
```

jfmengels:
> EDIT: It typechecks. But does it work at runtime? Need to check. EDIT2: Nope. Why is TypeScript so hard to get right with dependencies…

Yes, I went through the same cycle you just went through :sweat_smile:

lydell:
Oh well, I must have misunderstood something and made some mistake in elm-tooling. I’ve only tested it with pure JS (bot `require` and `import`) and .ts TypeScript with `import` before.

So maybe just go with `@ts-ignore` (or `@ts-expect-error`) for now. Maybe the PR getting rid of elm-tooling gets finished soon anyway…

lydell:
Found a workaround if you *really* want typechecking for those calls:

```js
const elmToolingCliImport = require("elm-tooling");
const getExecutableImport = require("elm-tooling/getExecutable");

const elmToolingCli = /** @type {typeof elmToolingCliImport["default"]} */ (
  /** @type {unknown} */ (elmToolingCliImport)
);

const getExecutable = /** @type {typeof getExecutableImport["default"]} */ (
  /** @type {unknown} */ (getExecutableImport)
);
```

jfmengels:
Thanks, I'll keep it as is for now I think, but happy to know that your comment will be here if I change my mind :smile:

Issue #125, opened 2023-05-05
Add types to JS files


jfmengels:
I have added TypeScript to the project to help out with finding bugs during development.

Unfortunately, because without type annotations it determines that the arguments of pretty much every function is `any`, it doesn't report many problems in practice.

The idea is to add more types, so that there are none of these implicits `any`.

I have added TypeScript configuration that enables that setting for a few files, and have added all the necessary annotations for them, but I'm still missing a lot of the files. You can run it using

```
npx tsc --project tsconfig.no-implicit-any.json --watch
```

The idea is to update the `tsconfig.no-implicit-any.json` by adding a new file in the `include` list one and adding types until there are no more issues, and repeating until all files are done and we can enable the `noImplicitAny` setting in the regular `tsconfig.json` file.
I recommend adding new files that only import files that are already in the `include` list. (To do that, just add an arbitrary file, and if TS reports issues for other files, then choose one of those files instead, and so on).

Note that adding type annotations and fixing issues can result in the "main" tsconfig.json starting to report new issues.

If you'd like to work on this, great! And thank you!
Feel free to do as much or as little as you wish, I recommend to do small steps in this work, and to do for instance one PR per file added.

lishaduck:
For the sake of future reference: #165, #166, #177, #179, #180, #181, #182, & #183.

lishaduck:
Status: 27/51 production JS files (53%)

lishaduck:
Making progress in #301! 

Status: 40/53 files

lishaduck:
48/53!

lishaduck:
> I highly appreciate the work and excitement, but go write your essay! 😅
> (as much as I understand why you prefer doing something else)

I know, I know. I'm gonna go finish it now 🙃
I do like writing, and I like the topic, this is just a lot more fun...

Well, off to write another couple hundred words...

lishaduck:
Oh, can you close #132 as superseded, @jfmengels?

PR #126, opened 2023-05-05
Add types to options.js

henriquecbuss:
This PR is a step towards #125, and adds types to `options.js`.

It also fixes a `@ts-expect-error` in the same file -- that's why some other files are touched.

Thanks for the project, glad I can help in some way 🙏

### DIFF
--- a/lib/error-message.js
+++ b/lib/error-message.js
@@ -44,7 +44,7 @@ function report(options, err, defaultPath) {
 
   return Anonymize.pathsAndVersions(
     options,
-    formatHuman(options?.debug ?? false, err)
+    formatHuman((options && options.debug) || false, err)
   );
 }
 
--- a/lib/error-message.js
+++ b/lib/error-message.js
@@ -22,13 +22,13 @@ class CustomError extends Error {
 }
 
 /**
- * @param {Options} options
+ * @param {Options | null} options
  * @param {CustomError} err
  * @param {Path} [defaultPath]
  * @returns {string}
  */
 function report(options, err, defaultPath) {
-  if (options.report === 'json') {
+  if (options && options.report === 'json') {
     return Anonymize.pathsAndVersions(
       options,
       Anonymize.pathsAndVersions(
@@ -42,7 +42,10 @@ function report(options, err, defaultPath) {
     );
   }
 
-  return Anonymize.pathsAndVersions(options, formatHuman(options.debug, err));
+  return Anonymize.pathsAndVersions(
+    options,
+    formatHuman(options?.debug ?? false, err)
+  );
 }
 
 /**
--- a/lib/main.js
+++ b/lib/main.js
@@ -49,20 +49,27 @@ function errorHandler(err) {
 async function runElmReview() {
   const {elmModulePath, reviewElmJson, appHash} = await Builder.build(options);
 
-  if (!elmModulePath) {
+  if (!elmModulePath || !options.elmJsonPath) {
     AppState.exitRequested(1);
     return;
   }
 
+  const optionsWithElmJsonPath = {...options, elmJsonPath: options.elmJsonPath};
+
   const [{app}] = await Promise.all([
-    Builder.buildElmParser(options, reviewElmJson).then(() =>
-      Runner.initializeApp(options, elmModulePath, reviewElmJson, appHash)
+    Builder.buildElmParser(optionsWithElmJsonPath, reviewElmJson).then(() =>
+      Runner.initializeApp(
+        optionsWithElmJsonPath,
+        elmModulePath,
+        reviewElmJson,
+        appHash
+      )
     ),
 
-    ResultCache.load(options, appHash)
+    ResultCache.load(optionsWithElmJsonPath, appHash)
   ]);
 
-  const success = await Runner.runReview(options, app);
+  const success = await Runner.runReview(optionsWithElmJsonPath, app);
   AppState.exitRequested(success ? 0 : 1);
 }
 
@@ -72,7 +79,7 @@ async function runElmReviewInWatchMode() {
     await Builder.build(options);
   await Builder.buildElmParser(options, reviewElmJson);
 
-  if (!elmModulePath) {
+  if (!elmModulePath || !options.elmJsonPath) {
     Watch.watchConfiguration(options, reviewElmJson, reviewElmJsonPath, () => {
       AppState.buildRestarted();
       runElmReviewInWatchMode();
@@ -80,8 +87,10 @@ async function runElmReviewInWatchMode() {
     return;
   }
 
+  const optionsWithElmJsonPath = {...options, elmJsonPath: options.elmJsonPath};
+
   const initialization = await Runner.initializeApp(
-    options,
+    optionsWithElmJsonPath,
     elmModulePath,
     reviewElmJson,
     appHash
--- a/lib/options.js
+++ b/lib/options.js
@@ -12,6 +12,7 @@ const ErrorMessage = require('./error-message');
  * @typedef { import("./types/options").Options } Options
  * @typedef { import("./types/options").Subcommand } Subcommand
  * @typedef { import("./types/options").Template } Template
+ * @typedef { import("./error-message").CustomError } CustomError
  */
 
 /*
@@ -604,7 +605,7 @@ ${Flags.buildFlag(subcommand, Flags.gitHubAuthFlag)}`
 
 /**
  *
- * @param {ErrorMessage.CustomError} errorToReport
+ * @param {CustomError} errorToReport
  */
 function reportErrorAndExit(errorToReport) {
   console.log(ErrorMessage.report(null, errorToReport));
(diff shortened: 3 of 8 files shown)


lib/error-message.js
```diff
@@ -22,13 +22,13 @@ class CustomError extends Error {
...
 /**
- * @param {Options} options
+ * @param {Options | null} options
  * @param {CustomError} err
  * @param {Path} [defaultPath]
  * @returns {string}
  */
 function report(options, err, defaultPath) {
-  if (options.report === 'json') {
+  if (options && options.report === 'json') {
```
jfmengels, line 31: I'm surprised, are there ever instances where options is null? 🤔

henriquecbuss: There was a comment to fix it later:

https://github.com/jfmengels/node-elm-review/blob/9fa20ed5cec3e237fa205bcfe7c714fe157d7600/lib/options.js#L543-L545

It wasn't null, but an empty object. Since the `reportErrorAndExit` function didn't get any `Options` as args, I thought `null` would be better than `{}`.

But maybe that wasn't what you meant to do! I can always revert back, make `report` accept only `Options` (as opposed to `Options | null`), and put the comment back

jfmengels: AH yes. It seems for that location I was pretty lazy back when I added it. This works for me I think :+1:

lib/error-message.js
```diff
@@ -42,7 +42,10 @@ function report(options, err, defaultPath) {
     );
   }
 
-  return Anonymize.pathsAndVersions(options, formatHuman(options.debug, err));
+  return Anonymize.pathsAndVersions(
+    options,
+    formatHuman(options?.debug ?? false, err)
```
jfmengels, line 47: The CLI still supports Node v10, and `?.` and `??` are not supported until v14 :/

lib/options.js
```diff
@@ -201,7 +204,7 @@ try re-running it with ${chalk.cyan('--elmjson <path-to-elm.json>')}.`,
     forceBuild: args['force-build'],
     report: args.report === 'json' || args.report === 'ndjson' ? 'json' : null,
     reportOnOneLine: args.report === 'ndjson',
-    rulesFilter: listOfStrings(args.rules),
+    rulesFilter: listOfStrings(args.rules) || [],
```
jfmengels, line 207: I think this should stay `null`, as I think we do check whether this has been set.
The Options and the different types I created in the `lib/types/*.d.ts` files have been created in a way to resemble what I did in JS, but might not always been correct. So far, the JS part seems way more correct than the types (it's for now the source of truth, though hopefully we'll find bugs with all of this!), so feel free to change the types if they seem to contradict JS.

https://github.com/jfmengels/node-elm-review/blob/df9a5107b01a8dcd8b56cc8988c10e39e7c94372/lib/result-cache.js#L83

lib/options.js
```diff
@@ -355,6 +380,10 @@ function findElmJsonPath(args, subcommand) {
   return findUp.sync('elm.json') || null;
 }
 
+/**
+ *
+ * @returns {(flag: string) => true}
```
jfmengels, line 385: Not saying you need to change it, but this is a smell that this should return `void` instead, right? :sweat_smile:

jfmengels (changes requested):
Thanks a lot for helping with this :pray:

lib/options.js
henriquecbuss: Oh yeah, for sure 😄

lib/options.js
henriquecbuss: Yes, there are checks for it everywhere it's used. I'll let it be null then 👍

henriquecbuss:
@jfmengels all of the comments should be addressed 🙂

lib/options.js
```diff
@@ -45,17 +45,25 @@ function compute(processArgv) {
     alias: Flags.flags
       .filter((flag) => flag.alias)
       .reduce((object, flag) => {
-        object[flag.name] = flag.alias;
+        if (flag.alias) {
+          object[flag.name] = flag.alias;
```
jfmengels, line 49: There has to be a way to tell TS that `flag.alias` is always false, right? :thinking:

lib/options.js
```diff
@@ -305,13 +323,19 @@ function listOfStrings(input) {
...
   }
 
   return null;
 }
 
+/**
+ *
+ * @param {Subcommand | null} subcommand
+ * @param {string} string
+ * @returns {{ repoName: string, pathToFolder: string, reference: string }}
```
jfmengels, line 337: This is `Template` from `options.d.ts`

jfmengels (commented):
This LGTM :+1: , but the `npm run tsc` test is failing, because these changes have introduced problems for the regular `tsconfig.json` configuration.

lib/options.js
henriquecbuss: Is this what you mean? https://github.com/total-typescript/ts-reset#make-filterboolean-filter-out-falsy-values

henriquecbuss:
There are a few errors in the regular `tsconfig.json` configuration, related to `options.elmJsonPath` being nullable. I'm not exactly sure how to go about this... does it make sense for it to be nullable? Is it even possible to run elm-review without an `elm.json`? Maybe we should error out if we can't find it?

jfmengels:
I can make sense for it be `null`, when you're running commands like `elm-review --help` or `init` for instance.
But for the review part, it should never be `null`. When it is, we will be throwing an error somewhere. I'm not sure what's the best way to prove that to TS. I guess casting could be an okay thing for now, until we figure out something better (if you don't want to scratch your head too much). Other ideas welcome, I'm not super familiar with TS.

lib/types/options.d.ts
```diff
@@ -60,6 +60,10 @@ export type Options = {
   gitHubPassword: string | undefined;
 };
 
+export type OptionsForRunningElmReview = Omit<Options, 'elmJsonPath'> & {
+  elmJsonPath: Path;
+};
+
```
henriquecbuss, line 66: If we think more types are going to have to change for running elm-review, we could also abstract this a little:

```ts
type OverriddenFieldsForRunningElmReview = { elmJsonPath: Path, ... }

export type OptionsForRunningElmReview = Omit<Options, keyof OverriddenFieldsForRunningElmReview> & OverriddenFieldsForRunningElmReview
```

Also, this name is pretty long, but I couldn't think of anything better 😅

lib/main.js
```diff
@@ -49,20 +49,27 @@ function errorHandler(err) {
 async function runElmReview() {
   const {elmModulePath, reviewElmJson, appHash} = await Builder.build(options);
 
-  if (!elmModulePath) {
+  if (!elmModulePath || !options.elmJsonPath) {
     AppState.exitRequested(1);
     return;
   }
 
+  const optionsWithElmJsonPath = {...options, elmJsonPath: options.elmJsonPath};
```
henriquecbuss, line 57: For now, we just exit if elmJsonPath is null. We could have better errors if you want (if so, we could wrap this in a function and just call it where needed) 🙂

lib/options.js
jfmengels: Yes, that's what I meant.  I don't know how smart `ts-reset` makes `filter` in this case, but I'd imagine that in this case TS will not remember that `flag.alias` is non-falsy. If it does, then yes, otherwise, it's so close :grin:

lib/options.js
```diff
@@ -540,9 +602,12 @@ ${Flags.buildFlag(subcommand, Flags.gitHubAuthFlag)}`
   return {gitHubUser, gitHubPassword};
 }
 
+/**
+ *
+ * @param {ErrorMessage.CustomError} errorToReport
```
jfmengels, line 607: I think you're missing the import for `CustomError`, I don't think this is in scope.

jfmengels (changes requested):
Sorry for long delay! I was on a much needed vacation break!

I think the PR looks pretty good in general. Looking at the tests (`npx jest` to run those locally), some behavior has changed.
1) The check for `elm.json` is done earlier than the check for bad flags (`unknownCheck`), and now that triggers different and less helpful errors than before.

2) Somehow the suppressions behave differently compared to before?

Sorry again for the delay, and I hope it's not too much of a bother (resuming) working on this. If it is, simply let me know and I'll try to get this across the finish line. I appreciate all of your help @henriquecbuss :heart: :blush:

henriquecbuss:
No worries! A much deserved break, for sure.

I'd love to keep working on this, but a lot of stuff (work + school) is going on at once, so I will not have time to continue working on this for a while (probably at least a month). It's fine if you want to finish this PR yourself, as that is quite a long time, but it's also fine if you want to wait for me to finish it (just know it will take a while). 

Hopefully I can contribute more as soon as I get more free time 🙂

lib/options.js
lishaduck: Fixed it locally, I tossed it into a gist:
https://gist.github.com/lishaduck/8cabda9ae913535f702a4376712d4608

I'd just give a patch, but I'm tweaking some other types. I'll PR soon (unless I forget 😉)

lishaduck:
Oh, @jfmengels, this can be closed. You made these changes already, the only changes left break the snapshots, as you observed. I plan on investigating it, but as the main purpose is complete I think it can be closed.

jfmengels:
Alright, thank you for looking into this @lishaduck. And thank you for the original work @henriquecbuss :pray:

PR #128, opened 2023-05-16
Add types to min-version.js

marc136:
As another tiny step towards https://github.com/jfmengels/node-elm-review/issues/125, this PR adds types to lib/min-version.js.


jfmengels:
Thank you! :pray:

Issue #82, opened 2022-09-09
Use fastest-levenshtein


jfmengels:
I see that our dependency https://www.npmjs.com/package/fast-levenshtein uses https://www.npmjs.com/package/fastest-levenshtein under the hood. Maybe that one would be enough to use use a dependency.

Maybe we could/should inline it as well?

marc136:
I had a short look at the code of `fast-levenshtein`, and replacing it with `fastest-levenshtein` should not be a problem.

elm-review uses `levenshtein.get(flag, f.name)` and does not pass the option to use locale-sensitive string comparisons. This means that `fast-levenshtein` does this:

```js
var levenshtein = require('fastest-levenshtein');
...
var Levenshtein = {
    get: function(str1, str2, options) {
        if (options.collate) { ... }
        return levenshtein.distance(str1, str2);
    }
    ...
}
```
See https://github.com/hiddentao/fast-levenshtein/blob/master/levenshtein.js#L84

Inlining `fastest-levenshtein` should pose no problem, because elm-review only needs 3 of the 4 functions (about 130 lines) and the code is MIT-licensed.

lishaduck:
Closed by #130.

PR #130, opened 2023-05-16
Replace fast-levenshtein with fastest-levenshtein

marc136:
This PR relates to #82 and removes one direct dependency and updates an indirect one.

fast-levenshtein@3.0.0 currently installs fastest-levenshtein@1.0.12 according to the package-lock.json.  
I updated to 1.0.16 because I saw no indication that the compiled typescript output should no longer be compatible to node.js 10.0.0.

Currently, elm-review uses `levenshtein.get(flag, f.name)` and does not pass the option to use locale-sensitive string comparisons. This means that `fast-levenshtein` does this:

```js
var levenshtein = require('fastest-levenshtein');
...
var Levenshtein = {
    get: function(str1, str2, options) {
        if (options.collate) { ... }
        return levenshtein.distance(str1, str2);
    }
    ...
}
```
See https://github.com/hiddentao/fast-levenshtein/blob/master/levenshtein.js#L84

I think it should not create a merge conflict with #126, but I'm not sure.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -2,7 +2,7 @@ const path = require('path');
 const chalk = require('chalk');
 const findUp = require('find-up');
 const minimist = require('minimist');
-const levenshtein = require('fast-levenshtein');
+const levenshtein = require('fastest-levenshtein');
 const packageJson = require('../package.json');
 const Flags = require('./flags');
 const ErrorMessage = require('./error-message');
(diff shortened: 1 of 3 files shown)


jfmengels:
Hey @marc136!
Sorry for the delay, I was on a much needed vacation :sweat_smile: :evergreen_tree: 
Thanks for looking into this! This looks good to me :blush:

marc136:
No worries, even with vacation, it was only a week 👟

PR #131, opened 2023-05-26
Extend the no-color check to include the NO_COLOR env var

qbit:
More info here: http://no-color.org/

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -1,8 +1,12 @@
 process.title = 'elm-review';
 
 if (!process.argv.includes('--no-color')) {
-  // Force `chalk` to add colors by default.
-  process.argv.push('--color');
+  if (process.env['NO_COLOR'] === undefined) {
+    // Force `chalk` to add colors by default.
+    process.argv.push('--color');
+  } else {
+    process.argv.push('--no-color');
+  }
 }
 
 const path = require('path');
--- a/lib/main.js
+++ b/lib/main.js
@@ -1,11 +1,11 @@
 process.title = 'elm-review';
 
 if (!process.argv.includes('--no-color')) {
-  if (process.env['NO_COLOR'] === undefined) {
+  if (process.env['NO_COLOR'] === "1") {
+    process.argv.push('--no-color');
+  } else {
     // Force `chalk` to add colors by default.
     process.argv.push('--color');
-  } else {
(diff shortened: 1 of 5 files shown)


jfmengels:
Hey @qbit! This sounds good to me. Have you tried running `elm-review` to see it work? Because it's not working for me :thinking:

qbit:
Oh crud. I tested the logic in an external file.. Let me re-work it.

qbit:
Should be good no. Sorry about that.

jfmengels:
Sorry for the delay in looking into this!
This will be in there with the next version (2.11.0), thank you for your work and suggestion :blush:

lib/main.js
```diff
@@ -1,11 +1,11 @@
 process.title = 'elm-review';
 
 if (!process.argv.includes('--no-color')) {
-  if (process.env['NO_COLOR'] === undefined) {
+  if (process.env['NO_COLOR'] === "1") {
```
lydell, line 4: https://no-color.org/ says:

> Command-line software which adds ANSI color to its output by default should check for a `NO_COLOR` environment variable that, _when present and not an empty string **(regardless of its value)**_, prevents the addition of ANSI color.

The next paragraph I’ve always interpreted as an _example:_

> By adopting this standard, users that prefer to have plain, non-colored text output can export `NO_COLOR=1` to their shell’s environment and automatically disable color by default in all supported software.

As far as I know, `NO_COLOR=0` also turns off color. It’s only based on the presence, not the value.

jfmengels: Gotcha, my bad. But doing `No_COLOR=""` should allow for colors (I tried this on the GitHub CLI, one of the projects they list in that file). I'll fix that.

jfmengels: Fixed on master (I amended the history a bit to remove confusion)

lydell: `NO_COLOR=""` is probably inconsistent between tools, likely due to the empty string being falsy in many languages.

In elm-watch and other programs I’ve made I just do `if ("NO_COLOR" in process.env)` because I interpreted the “spec” as if the `NO_COLOR` is set at all, regardless of value (including the empty string), it should mean “no colors”.

I don’t care about which way you go in elm-review though, just wanted to share this detail 😄

PR #138, opened 2023-09-09
Fix elm-tooling typings

lydell:
I finally figured out how to fix the TypeScript typings in the elm-tooling CLI: https://github.com/elm-tooling/elm-tooling-cli/blob/main/CHANGELOG.md#version-1141-2023-09-09

This PR updates to the latest version of elm-tooling (with the typings fix) and gets rid of the TypeScript disable comments which are no longer needed.


jfmengels:
Thank you! :pray:

PR #143, opened 2023-09-19
TCO for formatReports

lue-bird:
Like https://github.com/jfmengels/elm-review-reporter/pull/5 but with your suggestions applied.

> With regards to the solution, it is now TCO, but it will also be terrible slow (though looking at it, the current version is also probably not super fast... not sure which is faster), because we prepend the accumulated list N times (for N report items), and that's N² operations.

This was also the case for the current, non-TCO version but should be fixed in this PR using the method you suggested:

> Instead we probably want to do the classic newItem :: accumulator with a List.reverse at the very end, which is 2n operations.

### DIFF
--- a/template/src/Elm/Review/Reporter.elm
+++ b/template/src/Elm/Review/Reporter.elm
@@ -781,19 +781,34 @@ fixableErrors files =
 
 formatReports : DetailsMode -> Mode -> List FileWithError -> List Text
 formatReports detailsMode mode files =
+    formatReportsEndingWith [] detailsMode mode files
+
+
+formatReportsEndingWith : List (List Text) -> DetailsMode -> Mode -> List FileWithError -> List Text
+formatReportsEndingWith soFarReverse detailsMode mode files =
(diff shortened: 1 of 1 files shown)


jfmengels (approved):
Nice, thank you!
After this, I will try to see whether it would make sense to use Arrays instead of List for the accumulation. Maybe that's faster.

lue-bird:
In my experience, pushing to arrays is really not fast compared to `(::)`.
If any option is faster, it would be [minibill's elm-rope](https://dark.elm.dmy.fr/packages/miniBill/elm-rope/latest/) which applies nested concats in general in linear time.

jfmengels:
Yeah, `(::)` is really fast, but using `(::)` plenty of times and then reversing *might* be slower than using arrays (especially with the quite large number of concats need in this area). I'll have to benchmark this instead.
If it's faster, then that's good. Otherwise, I might take a look at `elm-rope` :eyes:

Issue #137, opened 2023-08-23
elm-review hangs in new project


wise-tim:
Follow up from https://elmlang.slack.com/archives/C010RT4D1PT/p1692789314256189

I ran elm-review in my Lamdera app on apple-silicon and it hangs with no output

Here is the output from elm-review --debug:

`elm-review --debug
Starting review application build
-- CONFIGURATION COMPILATION ERROR ---------------------------------------------

I encountered a problem when solving dependencies:

elm-json exited with signal SIGKILL

(empty stderr)
Error: I encountered a problem when solving dependencies:

elm-json exited with signal SIGKILL

(empty stderr)
    at /opt/homebrew/lib/node_modules/elm-review/lib/template-dependencies.js:50:13
    at /opt/homebrew/lib/node_modules/elm-review/lib/template-dependencies.js:291:29
    at async Object.getOrCompute (/opt/homebrew/lib/node_modules/elm-review/lib/cache.js:13:18)
    at async Promise.all (index 0)
    at async createTemplateProject (/opt/homebrew/lib/node_modules/elm-review/lib/build.js:315:43)
    at async Promise.all (index 1)
    at async buildLocalProject (/opt/homebrew/lib/node_modules/elm-review/lib/build.js:160:25)
    at async runElmReview (/opt/homebrew/lib/node_modules/elm-review/lib/main.js:50:51)`

Thanks for you help!

jfmengels:
Thank you for the issue!
Can you tell me which version of Mac OS you're running? Is it an older version maybe? I know that on Monterey and older there are some issues running elm-json.

Also, can you tell me what the version of elm-json is? If you installzd elm-review locally, then elm-json should be in your node_modules (either directly or under elm-review I believe).

timtimtam:
Apologies for the delay:

I am running Ventura 13.3 on an M1
I couldn't find elm-json anywhere in node-modules, but when I `npm i elm-json` I get version `0.2.13`

Running `npx elm-json tree` also hangs... which lead me to find this elm-json issue https://github.com/zwilias/elm-json/issues/54 that looks like it could describe my problem too. I think this, and the linked (but now expired) slack thread I what I recalled seeing earlier.

There is an open pull request that may fix it
https://github.com/zwilias/elm-json/pull/55/files

I might try and build elm-json from source to see if it works for me.

timtimtam:
Please feel free to close this issue if you think it is appropriate.

jfmengels:
[...]

PR #144, opened 2023-10-14
Add --offline flag and replace elm-json by elm-solve-deps-wasm

### LINKED ISSUE #51: Error: elm-json exited with signal SIGABRT
When using elm-review in the Nix package manager, I ran into this message, so am posting it here as requested:

(Sorry if I'm doing this wrong. I've never posted a Github issue before.)

```

  https://github.com/jfmengels/node-elm-review/issues/new

Please include this error message and as much detail as you can provide. If you
can, please provide a setup that makes it easy to reproduce the error. That will
make it much easier to fix the issue.

Error: elm-json exited with signal SIGABRT
    at ChildProcess.<anonymous> (/nix/store/04f2xpjs4z3qnz0qn4i018vb0wzggjmw-node_elm-review-2.4.6/lib/node_modules/elm-review/lib/template-dependencies.js:223:27)
    at ChildProcess.emit (events.js:375:28)
    at maybeClose (internal/child_process.js:1055:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:288:5)
```
These steps reproduce the issue (I am using Ubuntu Linux, but it should work on any Linux / Mac, where Nix is available):
1. install Nix package manager using [these instructions](https://nixos.org/guides/install-nix.html)
2. make a new directory and ```cd``` into it
3. make a new file called shell.nix with these contents:
```
{ pkgs ? import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/a165aeceda9f9741d15bc2488425daeb06c0707e.tar.gz") {} }:

pkgs.mkShell {
  buildInputs = [
    pkgs.elmPackages.elm-review
    pkgs.elmPackages.elm
  ];
}
```
4. run ```nix-shell --pure``` to go into an isolated shell
5. do ```elm init``` and agree to question
6. do ```elm-review init``` and agree to question

The issue only happens with the ```--pure``` option to nix-shell, so my theory is that elm-review is not properly packaged on Nix, and actually depends on something, probably elm-json from a quick look at the Javascript file mentioned in the error message.

### LINKED ISSUE #61: Consider allowing node-elm-review to work in offline mode
Firstly `elm-review` is a fantastic tool! Allowing for custom rules to align with team decisions is very liberating :) So thank you again for this!

I'd like to propose that there be an `offline` or `disconnected` type of execution mode that is intended primarily for CI. Although this is not a concern generally, within our CI configuration (hydra and nix) we build our code in a sandbox that does not have access to the internet. In this scenario, `elm-review` fails to execute due to `elm-json` being unable to communicate with `packages.elm-lang.com`.

In this case, using the offline execution, it is possible to short-circuit the `elm-json solve` step with the current contents of `elm.json`. There are, of course, caveats with this as the environment would be **expected** to have all dependencies present and available for use. This is a hard requirement for a sandbox build so not a problem in this scenario.

I'd be curious about your thoughts on this. A crude patch that we have been applying to `elm-review` locally is [available here](https://github.com/circuithub/node-elm-review/commit/d167d4548c72b50d25cad995eeeaf8d541e2be94) (currently based off of `elm-review` 2.5.1).

### LINKED ISSUE #81: Use elm-solve-deps-wasm instead of elm-json
This has been mentioned in other issues in this repository, like https://github.com/jfmengels/node-elm-review/issues/51 and https://github.com/jfmengels/node-elm-review/issues/61#issuecomment-1023483505

I think this has been shipped to `node-elm-test` (https://github.com/rtfeldman/node-test-runner/pull/558/) and I haven't heard problems about it.

@mpizenberg @lydell do you know whether `elm-solve-deps-wasm` is capable of doing all the things that `elm-review` currently requires `elm-json` for?

- Solving dependencies, like in https://github.com/jfmengels/node-elm-review/blob/a406f60c91790affdfc1f6cd31217c2fd02982e6/lib/template-dependencies.js#L40-L46
- Installing dependencies https://github.com/jfmengels/node-elm-review/blob/a406f60c91790affdfc1f6cd31217c2fd02982e6/lib/template-dependencies.js#L127-L144 (used for `elm-review init`). I guess that a way to get the latest compatible versions of packages would work as well.
- Upgrading dependencies https://github.com/jfmengels/node-elm-review/blob/a406f60c91790affdfc1f6cd31217c2fd02982e6/lib/template-dependencies.js#L150 This is used when using `--template` with a package that uses an out of date and incompatible `elm-review` version. To avoid asking every package author to update their configuration all the time, `elm-review` upgrades the version of the dependencies.

Is there anything else that you think I should know about this package?

Also, if one of you happens to be motivated by working on this, that would be amazing. I'm not sure when I would get to this myself.

jfmengels:
Fixes #61 by adding an offline mode
Fixes #81 by replacing `elm-json` by `elm-solve-deps-wasm`
Fixes #51 #137 by removing `elm-json` and all the related problems

This was not as straightforward as I thought, hence the mess of the commit history.
Turns out `elm-solve-deps` has a pretty different purpose than `elm-json`, but given enough tricks you can make it do the same thing.

Consider how many areas of the tool this change affects (regular review, init, new-package, --template, ...), I'm willing to release this as a beta version so that people can try it out and report problems before we cut a new release.

@gpampara I've added a `--offline` flag. Do you think you need a `elm-review build` (name TBD) command to pre-compile the application (into `elm-stuff` + `ELM_HOME`) so that you can then run it in `--offline` mode in CI?

Remaining:
- [x] Document `--offline` flag
- [x] Verify all places that make HTTP requests are disabled through `--offline` in review mode
- [x] Add a command to prepare elm-review to run offline
- [x] Add nice error message when run in `--offline` mode and/or dependencies could not be figured out
- [x] Make sure we don't use versions of packages that the configuration didn't specify. (Just move the dependencies from indirect to direct dependencies?)
- [x] When initializing a project using a template, make sure test dependencies are available and updated


gpampara:
@jfmengels This is great! I think an option like that might be quite useful and I'd like to hear the comments of others on this. I don't foresee that the build needs to happen ahead of time, but all dependencies would be needed for this offline operation.

jfmengels:
@gpampara How do you currently download all dependencies?

gpampara:
@jfmengels We use the tool [elm2nix](https://github.com/cachix/elm2nix) to prepare all the dependencies into Nix, but it's essentially means that upon build they are available. This is then used to recreate the `ELM_HOME` locally such that when the compiler runs, it's none the wiser about how the dependencies got there and because they are there at the correct versions, compile runs without trying to go online. This would be similar to caching `ELM_HOME` on other CI systems (if you squint at it 😄)

PR #145, opened 2023-10-25
Avoid using elm-tooling to generate its JSON file

jfmengels:
Removes `elm-tooling` as a dependency since it's (now?) only used to create the `elm-tooling.json` in the `new-package` command. Since this file rarely changes, so @lydell suggested we remove the dependency and hardcode the `elm-tooling.json` file.

I added a test to keep the generated file somewhat up to date. It's not the best test, but it's something already.

cc @lydell if you have some additional remarks or suggestions :)


lydell (approved):
Looks great!

PR #147, opened 2024-03-17
Remove fs-extra

### LINKED ISSUE #120: Replace fs-extra
In https://github.com/jfmengels/node-elm-review/pull/119 we replaced most usages of `fs-extra`.
We kept it because of the copy capabilities that it has that are not in `fs` or `graceful-fs`.

(Note: it's fine to keep it in `new-package/maintenance/update-examples-from-preview.js`, that's a separate generated project for packages)

jfmengels:
Fixes #120

Removes the dependency on `fs-extra`. As @lydell pointed out, `fs.copyFileSync` already exists in core `fs`.
`fs.copy` (which copies entire directories) however doesn't. I chose to fix this by hardcoding which files to copy (which is faster, though potentially error-prone :shrug:)

Now that `fs-extra` is absent from the dependencies, `tsc` complains about the files that reference it in the `new-package/` folder. Those files are copied over and not included, but do use `fs-extra` (mostly because of the "need" to copy entire directories, ideally in a sync manner). I don't know how to suppress this error, and if I ignore the file, then ESLint complains that it can't parse it.

If anyone has an idea on how to fix this, I'm all ears!

### DIFF
--- a/lib/fs-wrapper.js
+++ b/lib/fs-wrapper.js
@@ -1,6 +1,9 @@
+const path = require('path');
 const util = require('util');
 const fs = require('graceful-fs');
 const {rimraf} = require('rimraf');
+
+const copyFile = util.promisify(fs.copyFile);
 
 // TODO When we drop support for Node.js v10, use natively promisified functions
 
@@ -87,6 +90,20 @@ function mkdirpSync(dir) {
   fs.mkdirSync(dir, {recursive: true});
 }
 
+/**
+ * Copy files from srcFolder to destFolder.
+ * @param {Path} srcFolder
+ * @param {Path} destFolder
+ * @param {Path[]} files
+ * @returns {Promise<void>}
+ */
+function copyFiles(srcFolder, destFolder, files) {
+  const promises = files.map((file) =>
+    copyFile(path.join(srcFolder, file), path.join(destFolder, file))
+  );
+  return Promise.all(promises).then(() => undefined);
+}
+
 module.exports = {
   readFile,
   readJsonFile,
@@ -97,6 +114,8 @@ module.exports = {
 
   mkdirp,
   mkdirpSync,
+
+  copyFiles,
 
   readdir: util.promisify(fs.readdir),
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Removed the dependency on `fs-extra`.
+
 ## [2.11.1] - 2024-03-16
 
 - Fixed a crash when running in watch mode related to not being able to fetch cache results.
--- a/lib/init.js
+++ b/lib/init.js
@@ -1,6 +1,5 @@
 const fs = require('fs');
 const path = require('path');
-const fsExtra = require('fs-extra');
 const chalk = require('chalk');
 const prompts = require('prompts');
 const FS = require('./fs-wrapper');
@@ -155,7 +154,7 @@ async function createElmJson(options, directory) {
 }
 
 function createReviewConfig(directory, template) {
-  fsExtra.copyFileSync(
+  fs.copyFileSync(
     path.join(__dirname, '../init-templates/', template),
     path.join(directory, 'ReviewConfig.elm')
   );
--- a/lib/new-package.js
+++ b/lib/new-package.js
@@ -3,7 +3,6 @@ const path = require('path');
 const childProcess = require('child_process');
 const chalk = require('chalk');
 const prompts = require('prompts');
-const fsExtra = require('fs-extra');
 const Init = require('./init');
 const Spinner = require('./spinner');
 const NewRule = require('./new-rule');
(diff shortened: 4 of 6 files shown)


lydell:
You can use `fs.cpSync(src, dest, { recursive: true })` to copy whole dirs. https://nodejs.org/api/fs.html#fscpsyncsrc-dest-options

Available since 16.7.0, marked as Experimental, though. I use it in scripts without trouble, but I’m not sure if I’d dare doing it in package code.

lib/fs-wrapper.js
```diff
@@ -1,7 +1,10 @@
+const path = require('path');
 const util = require('util');
 const fs = require('graceful-fs');
 const {rimraf} = require('rimraf');
 
+const copyFile = util.promisify(fs.copyFile);
```
lydell, line 6: Btw, there’s a a promises API: https://nodejs.org/api/fs.html#promises-api

jfmengels:
For now `elm-review` supports v10 of Node.js and above, although I think that support for v10 and v12 is broken because of indirect dependencies :disappointed:, but supporting older versions including v14 is still the aim.

Maybe I should break this in a future version because for a soon-to-be feature I'll need a more recent version of `glob` which only supports ">=16 || 14 >=14.17". Maybe that would be a good time to cut a major version of the CLI :shrug:

lib/fs-wrapper.js
jfmengels: Ah yes, I looked for it in my code but didn't see any usages, so I thought they must have appeared after v10.

lishaduck:
> You are correct, Node.js 10 and 12 are currently broken.

You actually need Node 16.14+ to resolve correctly:
https://github.com/jfmengels/node-elm-review/blob/e1958183e2e3355581296c1f865f7ef39bfa340c/package-lock.json#L10467
https://github.com/isaacs/node-lru-cache/blob/v9.0.0/package.json#L80-L82

On my Windows machine (my main Elm machine), I only have 16.12, so I've just had to ignore it for (two?) years. Luckily, it works, but npm complains mightily. I never thought to check if it what caused it,  otherwise I would have noted it sooner :)

---

Thoughts I wrote down relating to majors, that none need read, but which people could, should they wish to, for some unknown reason:

> - Node.js 10 and 12 have been EOL for quite some time now.
> - You can release major versions of the CLI without causing problems like with the Elm package (which would make all rule packages incompatible).

If you're dropping node versions, drop to 18+, and just tell folks below that to use npm:elm-review@2, assuming elm releases are decoupled (so everything still works). After all, Node 18 is still maintained, but 16 [uses an outdated OpenSSL, so it's out of support](https://nodejs.org/en/blog/announcements/nodejs16-eol). Then again, I don't like losing the 2-2 though, as it helps me mentalize it (is it `elm-pages` v3 or v10? 😉).
If you made a major with the elm package though, that might also be acceptable for people (i.e., breaking everywhere). On the other hand, that's majorly breaking to the ecosystem, but you could make a codemod, so maybe not terrible? Especially with @elm-review-bot?
Oh, and this is a good read by the creator of SemVer himself: [Major Version Numbers are Not Sacred](https://tom.preston-werner.com/2022/05/23/major-version-numbers-are-not-sacred).

lib/fs-wrapper.js
lishaduck: From the page above:

> Version | Changes
> -- | --
> v14.0.0 | Exposed as require('fs/promises').
> v11.14.0, v10.17.0 | This API is no longer experimental.
> v10.1.0 | The API is accessible via require('fs').promises only.
> v10.0.0 | Added in: v10.0.0

So you can use it in 10+, but it'll be experimental, and getting at it might be annoying.

lib/fs-wrapper.js
```diff
@@ -87,6 +90,20 @@ function mkdirpSync(dir) {
...
+ * @param {Path} srcFolder
+ * @param {Path} destFolder
+ * @param {Path[]} files
+ * @returns {Promise<void>}
+ */
+function copyFiles(srcFolder, destFolder, files) {
+  const promises = files.map((file) =>
+    copyFile(path.join(srcFolder, file), path.join(destFolder, file))
+  );
+  return Promise.all(promises).then(() => undefined);
```
lishaduck, line 104: Why not this?

```suggestion
  return Promise.all(promises);
```

Maybe I've forgotten something, but the noop shouldn't be necessary.

jfmengels: I think it's to have the return type `@returns {Promise<void>}` instead of `@returns {Promise<void[]>}` or `@returns {Promise<something[]>}`

lishaduck: > I think it's to have the return type `@returns {Promise<void>}` instead of `@returns {Promise<void[]>}` or `@returns {Promise<something[]>}`

Hmm. That makes sense. I would think that `void[]` is assignable to `void`, but maybe not.

jfmengels:
Supersed by #303

PR #150, opened 2024-05-02
Allow relative AND absolute paths when ignoring directories

### LINKED ISSUE #149: Allow relative AND absolute paths when ignoring directories
As discussed in the [slack thread](https://elmlang.slack.com/archives/C010RT4D1PT/p1714655201263869), it would be nice to be able to handle absolute and relative paths via the `--ignore-dirs` flag.

jamesrweb:
Fixes #149.

Convert absolute paths to be relative to the project being reviewed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- The `--ignore-dirs` and `--ignore-files` flags now support absolute paths. Thanks [@jamesrweb](https://github.com/jamesrweb)!
+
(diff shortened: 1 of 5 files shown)


jfmengels:
Thank you @jamesrweb :blush: 

I had to reverse the arguments to `path.relative`, but apart from that it seems to work well :relaxed:

jamesrweb:
> Thank you @jamesrweb :blush: 
> 
> 
> 
> I had to reverse the arguments to `path.relative`, but apart from that it seems to work well :relaxed: 

Nice! 🚀

PR #151, opened 2024-05-02
fix process.exit calls

BrianHicks:
This replaces all instances of `process.exit` with a new vendored dependency `exit` that drains all streams before exiting for real.

Fixes [the issue we talked about in the Elm slack](https://elmlang.slack.com/archives/C010RT4D1PT/p1714676525378159). For when that gets eaten by the Slack history monster, the problem is that `process.exit` would cut off JSON output when run from within a Node parent process.


jfmengels:
Seems to work really well. Thank you very much :heart: :pray:

PR #155, opened 2024-06-15
Add types to `app-wrapper.js`

lishaduck:
I'm unsure as what type `flags` is supposed to be, beyond that it's not `Flags`.
I'm pretty sure VSCode was wrong that it's `Options`.

_**Relates to:** #125_

### DIFF
--- a/lib/app-wrapper.js
+++ b/lib/app-wrapper.js
@@ -2,12 +2,40 @@ const path = require('path');
 const ElmCommunication = require('./elm-communication');
 const loadCompiledElmApp = require('./load-compiled-app');
 
+/**
+ * @typedef { import('./types/options').Options } Options
+ * @typedef { import('./types/options').ReviewOptions } ReviewOptions
+ * @typedef { import('./types/app').App } App
+ * @typedef { import('./types/app-wrapper').Exports } Exports
+ */
+
+/**
+ * // `@template` should be @typeParam, but TS is acting up.
+ *
+ * @template T
+ * @typedef { import('./types/promisify-port').CallbackFn<T> } CallbackFn
+ */
+
+/**
+ * // `@template` should be @typeParam, but TS is acting up.
+ *
+ * @template T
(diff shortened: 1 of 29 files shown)


lib/app-wrapper.js
```diff
@@ -2,12 +2,25 @@ const path = require('path');
 const ElmCommunication = require('./elm-communication');
 const loadCompiledElmApp = require('./load-compiled-app');
 
+/**
+ * @typedef { import('./types/options').Options } Options
+ * @typedef { import('./types/options').ReviewOptions } ReviewOptions
+ * @typedef { import('./types/app').App } App
+ */
```
lishaduck, line 9: Once ts 5.5 is stable (which should be soon), we can switch to the import syntax instead here:
```suggestion
/**
 * @import {Options} from './types/options';
 * @import {ReviewOptions} from './types/options';
 * @import {App} from './types/app';
 */
```

lishaduck:
Oh, I see. This is blocked on #126. I'll try to see if I fix it sometime.

lishaduck:
Ok, the failures from #126 were a bit intimidating, so I figure I'll just port them over here and just coauthor them.

lishaduck:
I plan on relanding this as separate PRs later this week.

jfmengels:
> I'm unsure as what type flags is supposed to be, beyond that it's not Flags.
I'm pretty sure VSCode was wrong that it's Options.

Flags is something that looks very much like `Options`. If I'm not mistaken, it's an object that is passed to the `worker_threads` that takes a number of fields (with some differences) taken from `Options`.

lib/app-wrapper.js
lishaduck: > soon

Today!
I'm currently cleaning up this branch locally, I should have something up today or tomorrow (certainly tomorrow in Europe. How's the future? 😉).

lishaduck:
Yikes! I'm up to 17 commits on my branch. I really need to split them up 😬

Issue #133, opened 2023-07-06
Update Node engine version in package.json to `>=16.14`


hingew:
### Description 

The minimal node version should be `16.14.0`. In a ci pipline with a lower version we got the following error: 

```
$ npm ci --cache .npm --prefer-offline
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'lru-cache@10.0.0',
npm WARN EBADENGINE   required: { node: '14 || >=16.14' },
npm WARN EBADENGINE   current: { node: 'v16.13.1', npm: '8.1.2' }
npm WARN EBADENGINE }
```

Dependency tree for `lru-cache`
```
-> npm ls lru-cache
├─┬ elm-review@2.10.2
│ └─┬ rimraf@5.0.1
│   └─┬ glob@10.3.1
│     └─┬ path-scurry@1.10.0
│       └── lru-cache@10.0.0
```

According the docs from `lru-cache` (https://github.com/isaacs/node-lru-cache#breaking-changes-in-version-8) the engine version should be `16.14.0`  or higher. 

Tnks for elm-review :heart:.

lishaduck:
https://github.com/jfmengels/node-elm-review/blob/e1958183e2e3355581296c1f865f7ef39bfa340c/package-lock.json#L10467
https://github.com/isaacs/node-lru-cache/blob/v9.0.0/package.json#L80-L82

On my Windows machine (my main Elm machine), I only have 16.12, so I've just had to ignore it for (two?) years. Luckily, it works, but npm complains mightily. I never thought to check if it what caused it,  otherwise I would have noted it sooner :)

_Originally posted by @lishaduck in https://github.com/jfmengels/node-elm-review/issues/147#issuecomment-2169064026_

lishaduck:
Did some more investigation:

lru-cache 10 re-adds support for Node 14, but, as @hingew noted, this `path-scurry` requires v9.
Latest `path-scurry` uses `lru-cache` v10, and latest `glob` uses latest `path-scurry`. So far, so good.
Latest `rimraf` supports `engines` `>=14.18`, so that too seems fine.


You know, maybe we should just refresh our lockfiles and pray it gets fixed. It seems fine, so it's either stale or a bug in npm.

jfmengels:
[...]

Issue #118, opened 2023-04-17
CI keeps failing


jfmengels:
THe CI keeps failing, even though the tests pass locally. This reduces my trust in the CI, and may at some point cause me to ship a version with broken code...

hingew:
I just ran into the same issue. I tested the tests locally with multiple different node versions:

| Version | Result |
| ------- | ------ | 
| 14.17.6 | :x: |
| 16.20.2 | :x: |
| 18.20.3 | :x: |
| 20.14.0 | :heavy_check_mark: |
| 22.3.0  | :x: |

I guess the snapshots for the tests where recorded with the node version `20.x` but in the CI we use the version `16.x`. The node version `16.x` and `20.x` gives diffrent outputs for `JSON.parse(...)` errors.

A possible solution for this would be to enforce a node version for devleopment. This could be archived with one of the following options

-  [asdf](https://github.com/asdf-vm/asdf) and a `.tool-versions` file 
-  [nvm](https://github.com/nvm-sh/nvm) and a `.nvmrc`

jfmengels:
Fixed by #157

PR #157, opened 2024-06-17
Nodejs version enforcement for development

hingew:
Issues: #133  #118 

### Development node version

This PR will suggest to use the node version 16 for development. This can be archived with `asdf` or `nvm`  (see #118 for details). 

### Runtime node version

The minimum runtime version, determined through `ls-engines`, is `>=14.17`.
This PR also adds the use of `ls-engines`, as suggested from @jfmengels (#133). With this PR the `ls-engines` runs in the CI to make sure that we match the engines we have specified in the `package.json`.


jfmengels:
@SocketSecurity ignore npm/array.prototype.some@1.1.6

jfmengels:
Thank you very much @hingew. I hope sticking to v16 won't be a problem for most people. Otherwise, we can always choose a newer version later.

The only thing we should add is add some guarantees that we don't use features from Node.js that are not available in v14.

This fixes #157 and #118.

lishaduck:
> The only thing we should add is add some guarantees that we don't use features from Node.js that are not available in v14.

I'm bumping devdeps right now, and eslint-plugin-n supports this, which you already use.

Issue #156, opened 2024-06-15
DevEx improvements


lishaduck:
Hi, @jfmengels! First off, thanks for elm-review! It's super useful and I use it whenever I write Elm. It's just great.

I've been getting used to the codebase, and I had a few thoughts:
1) Use `turbo` instead of `npm run`. Turbo is a faster, parallel task runner written in Rust by the Vercel team. It'd help speed up `npm run test`, which is pretty slow on my machine.
2) Author `.ts`, not `.d.ts`. I respect that you don't want to write this in TypeScript, that's completely fair! However, `.d.ts` is a compiled output format that shouldn't be manually written, [per the typescript team](https://github.com/microsoft/TypeScript/wiki/FAQ#:~:text=In%20general%20it%27s%20not%20correct%20to%20have%20a%20.d.ts%20file%20be%20a%20user%2Dauthored%20input%20to%20your%20program%20%2D%20even%20if%20a%20file%20only%20contains%20types%2C%20it%20should%20still%20be%20.ts%20so%20that%20it%20produces%20a%20.d.ts%20in%20the%20output%20for%20downstream%20consumers%20to%20use.). I'd switch the extensions (just for the types, not the source.[^1])
3) Switch to vitest—it's faster than Jest, and (as far as I can tell) supports CJS (though I've only authored ESM before so don't quote me).

I'd be happy to do all this, 1 and 2 are trivial and are directly annoying me, so I'm motivated 😄

[^1]: If you're fine with bumping _development_ Node to `^18.18.0 || >=20.6.0`, `tsx` can run TS natively, like `ts-node` but better. It would make predicates format a lot nicer, but that's a pretty weak reason so I don't think it's necessary.

lishaduck:
Another thing: I hit GH rate limits bisecting tests and now I can't run them at all.

jfmengels:
Hi @lishaduck!

Thanks for all the PRs! I'll be away at Elm Camp this week, I'll try to review them when I can, sorry if there's a bit of delay in handling them!

> 1. Use turbo instead of npm run. Turbo is a faster, parallel task runner written in Rust by the Vercel team. It'd help speed up npm run test, which is pretty slow on my machine.

In some projects, I use `npm-run-all` to run things in parallel. One downside with it is that if you do run things in parallel and one fails, which one failed can be a bit unclear. Anyway, I've added `npm run test:parallel` to do that. That said, there are some very slow and long tests, so they will keep taking a lot of time.

[...]

PR #164, opened 2024-06-22
Turbo-ize

lishaduck:
Depends on #163
Relates to #156

npm-run-all kept acting up, so I went ahead and migrated elm-review to turbo.
Key things: everything is parallelized now. It's also cached.

To run tasks, use `turbo ...` instead of `npm run ...`, which is installable via `npm install -g turbo`
You can also use `npx turbo`. They both run using _our_ pinned turbo, the global one just proxies to the local one.
And, of course, you can still run `npm run ...`, it just won't get cached.

If you want to add a task, just add it in the package.json. It'll automatically get cached. If you find it gets invalidated too much, you can add it to `turbo.json`.

Here's the diff as GH doesn't play well with stacks from forks: <https://github.com/lishaduck/node-elm-review/compare/bump-deps-2...turbo>
I also updated CI. I hope you don't mind.


package.json
```diff
@@ -9,16 +9,13 @@
   "packageManager": "npm@10.7.0",
   "scripts": {
     "elm-format": "elm-format --validate ast-codec init-templates new-package parseElm review template",
-    "test": "npm-run-all --print-name --silent elm-tests elm-format tsc eslint-check prettier-check test-run",
```
jfmengels, line 10: Do you know if it's possible to keep a test script so that I can just run `npm test` ?

jfmengels (approved):
Thank you for the PR. I'm willing to try `turbo` out.

Can you tell me where I can find the docs for turbo? I'm having trouble finding it.

Do you also know if there's a way to not have the interactive window showing up? Mostly out of curiosity.

> npm-run-all kept acting up

Do you remember what kind of problems you were encountering?

lishaduck:
> ~Can you tell me where I can find the docs for turbo? I'm having trouble finding it.~ It's at https://turbo.build/repo/docs I kept getting confused because I though `turbo` was a different tool than `turborepo`.

Yeah. turbo is the cli, which came first. Then they made turbopack, so they retroactively named it turborepo, but they share a cli (well, they will eventually).

> Do you also know if there's a way to not have the interactive window showing up? Mostly out of curiosity.

The `--log-order` flag. You can pass either --log-order=grouped or --log-order=stream.
stream is essentially npm-run-all

> > npm-run-all kept acting up
> 
> Do you remember what kind of problems you were encountering?

It wasn't injecting node_modules/.bin into the path, so some tests were failing.

> Do you know if it's possible to keep a test script so that I can just run `npm test` ?

Yeah. I can rename `test` to, say, `testing`, and then alias `npm test` to `turbo testing`. It's meant to be used in a monorepo where you do that for all your tasks in your root, then use turbo in the individual packages, but it works well in single package repos as well.

jfmengels:
> I can rename test to, say, testing, and then alias npm test to turbo testing

I think this would be great. `npm test` (or actually `npm t`) is a very ingrained shortcut for me :smile:

lishaduck:
If you prefer the log-order flags, I just found out you can set an env variable:
Set `export TURBO_UI=0` and `export TURBO_LOG_ORDER=streamed` (or `grouped`) in your `.zshrc` (or `.bash_profile` or whatever)

PR #165, opened 2024-06-22
JSDoc Updates

lishaduck:
Sort, format, lint jsdocs.

Depends on #164

Here's the diff as GH doesn't play well with stacks from forks: https://github.com/lishaduck/node-elm-review/compare/turbo...jsdocs


lib/anonymize.js
```diff
@@ -7,8 +7,8 @@ machines, and also the same output when only the CLI version changes.
 const path = require('path');
 
 /**
- * @typedef { import("./types/options").Options } Options
- * @typedef { import("./types/path").Path } Path
+ * @import {Options} from "./types/options"
```
jfmengels, line 10: Yay for TS 5.5 :tada:

vendor/node-elm-compiler.js
```diff
@@ -1,9 +1,5 @@
 'use strict';
 
-/**
- * @typedef { import("../lib/types/options").Options } Options
```
jfmengels, line 4: Is there a tool out there that you know of that can help finding these unused imports? Or do we already have one?

jfmengels:
Thanks again. There was a TS error that was introduced here but I pushed a fix.

vendor/node-elm-compiler.js
lishaduck: TS can do it, but it's buggy, so I turned it back off after I turned it on.
microsoft/TypeScript#58969

lishaduck:
> Thanks again ☺️ There was a TS error that was introduced here but I pushed a fix.

Thanks! I did a bunch of rebasing and must've missed one.

jfmengels:
Yeah, I jsut noticed you fixed the issue in the typings PR. That's fine, no worries :+1:

PR #167, opened 2024-06-22
Sync Support

lishaduck:
Sync support with jfmengels/elm-review#174
Also, fix a link in the readme.

---

As a random human, do you mind deleting `node-elm-review:master`?
It's confusing. As is using `elm-review:master` rather than `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ elm-review init --template jfmengels/elm-review-config/package
 ```
 
 Once you have set up an initial configuration, you can add new rules. As `elm-review` does not
-[come with built-in rules](https://github.com/jfmengels/elm-review/blob/main/documentation/design/no-built-in-rules.md),
+[come with built-in rules](https://github.com/jfmengels/elm-review/blob/master/documentation/design/no-built-in-rules.md),
 you can look for packages with rules on the [Elm package registry](https://package.elm-lang.org/) by searching for packages named `elm-review-`.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -112,7 +112,7 @@ elm-review init --template jfmengels/elm-review-config/package
 ```
 
 Once you have set up an initial configuration, you can add new rules. As `elm-review` does not
-[come with built-in rules](https://github.com/jfmengels/elm-review/blob/main/documentation/design/no-built-in-rules.md),
+[come with built-in rules](https://github.com/jfmengels/elm-review/blob/master/documentation/design/no-built-in-rules.md),
```
jfmengels, line 115: > As a random human, do you mind deleting node-elm-review:master?
It's confusing. As is using elm-review:master rather than main.

Yes, I can do that. But then this link should not get updated, correct?

jfmengels:
Sorry for the confusion between master and main! :disappointed: 
I've removed it now. I'll pull the fix from this PR without the link change from main to master.
Let me know if I was mistaken to do so.

lishaduck:
> Sorry for the confusion between master and main! 😞
> I've removed it now. I'll pull the fix from this PR without the link change from main to master.
> Let me know if I was mistaken to do so.

This link fix is still valid. `elm-review` uses `master`, not `main`. It would probably be less confusing to also rename it there as well.

jfmengels:
Oh right, that was indeed confusing. Ok, well time for me to change it over there!

PR #168, opened 2024-06-22
ESLint

lishaduck:
This is very much a WIP.
I'm working on modernizing the linting configuration now that node 10&12 are dropped.

I have a few stylistic questions, but first I want to get tests passing.


lishaduck:
Oh, tests pass now :shrug:
Whatever jest. 😆

lishaduck:
May I request a re-run of CI? It seems like a flake. It passed a moment ago, and the force-push shows it's clearly exactly the same code, just based off main so I could read my git client.

lishaduck:
Also, this is going to conflict with #147. Do you want me to see how much I can port from that PR?

jfmengels:
@lishaduck That sounds good to me. I'm not sure what to do about that PR at the moment (I'm also not sure as to the state I left it in to be honest). But feel free to postpone that porting to after this gets merged, we can sort the conflicts later :man_shrugging:

lishaduck:
Ok! ESLint v9 doesn't support Node 16, so we're still on 8.
ESLint plugins have followed suit, so we're lagging a few versions behind XO, n, and Unicorn.
Otherwise, I think it's all up to date.

PS: I know you don't like expiring todos (and I agree with your logic!), but I think these are fine, given how they work.

lishaduck:
~~This should be good to go!~~ Nope. Nonetheless:

Follow ups:
- Reland the actual stuff from #155
- Get #147 merged in some form
- Enable [`prefer-await-to-then`](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-await-to-then.md)?
- Add [`eslint-plugin-perfectionist`](https://github.com/azat-io/eslint-plugin-perfectionist)?
- [Other](https://github.com/xojs/xo/blob/ee9f0a3d72d55df098fc321c4d54a1ea3804e226/package.json#L71) [rules](https://github.com/xojs/xo/blob/ee9f0a3d72d55df098fc321c4d54a1ea3804e226/package.json#L73) that the XO config doesn't include b/c peerdeps were funny.

lishaduck:
Moving this back to WIP.
Those flakes were updated snapshots, now I have to go back through everything.
Also, I realized that the XO cli does a lot more than the config, so I'll fun with that :)
Do you mind if I make a TODO list issue? Both so I remember what needs to get done, but also so you can see what I'm working on?

jfmengels:
> Moving this back to WIP.

Sure. Let's focus on getting it merged again if that's alright with you (and do unrelated work to that in a separate PR).

> Do you mind if I make a TODO list issue?

Sure, go ahead.

> Enable [prefer-await-to-then](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-await-to-then.md?rgh-link-date=2024-06-22T18%3A12%3A50Z)?

I'd say not, I find that `.then` does provide some purpose even in the presence of `await`. For instance, with the rule you can't write this
```js
promises = await Promise.all([
    someFunction(data).then(doSomething),
    someOtherPromise,
    ...
]);
```

which can be useful if you want to do more things in parallel. Or just `a = await someFunction(data).then(doSomething)` to avoid intermediate variable declarations.

> I realized that the XO cli does a lot more than the config, so I'll fun with that :)

Yes, it does a bit more, though I don't remember what it did. To be honest, I don't think we need to add too many ESLint rules and other tools (but feel free to suggest them!), I mostly want to help avoid regressions, and get some more guarantees here and there so that this project is maintainable and bug-free.

So far (surprisingly), the migration to TS has not caught any noticeable bugs, so the investment has been quite poor. Not that we shouldn't continue (especially if it's a bit fun for some people :smile:). The XO rules can definitely wait I'd say (although, I'm not sure what has now been disabled).

lishaduck:
> > Moving this back to WIP.
> 
> Sure. Let's focus on getting it merged again if that's alright with you (and do unrelated work to that in a separate PR).

Yeah. I just want to fix tests and the lints I realized weren't getting applied right.

> > Do you mind if I make a TODO list issue?
> 
> Sure, go ahead.

👍

> > Enable [prefer-await-to-then](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-await-to-then.md?rgh-link-date=2024-06-22T18%3A12%3A50Z)?
> 
> I'd say not, I find that `.then` does provide some purpose even in the presence of `await`. For instance, with the rule you can't write this
> 
> ```js
> promises = await Promise.all([
>     someFunction(data).then(doSomething),
>     someOtherPromise,
>     ...
> ]);
> ```
> 
> which can be useful if you want to do more things in parallel. Or just `a = await someFunction(data).then(doSomething)` to avoid intermediate variable declarations.

Yeah. I tried turning it on but ran into that, so I wasn't sure.

> > I realized that the XO cli does a lot more than the config, so I'll fun with that :)
> 
> Yes, it does a bit more, though I don't remember what it did. To be honest, I don't think we need to add too many ESLint rules and other tools (but feel free to suggest them!), I mostly want to help avoid regressions, and get some more guarantees here and there so that this project is maintainable and bug-free.

Essentially, it applies the plugins, so migrating to the config turned off a _lot_ of important rules, like `eslint-plugin-n` (with `eslint-plugin-es-x`), for example.

> So far (surprisingly), the migration to TS has not caught any noticeable bugs, so the investment has been quite poor. Not that we shouldn't continue (especially if it's a bit fun for some people 😄).

It's a lot of fun :)
Once this lot is finished, I plan on finished ts-ifying the rest.
It's certainly helpful to the extent of maintenance, as TSEslint 6 caught a good deal. It's still local for now until I finish unicorn and n, but it's definitely a good investment. And I turned off the abusive promises rules! Lots of stuff to fix (though mostly modernization and theoretical data races, not actual bugs).

> The XO rules can definitely wait I'd say (although, I'm not sure what has now been disabled).

Unicorn was off, but I've almost finished reenabling it. N needs to get on. We don't use AVA, so that can remain off.

lishaduck:
Ooooh. Unicorn found a unicode bug!
Probably not a real issue, but `String.prototype.split()` breaks apart surrogate pairs.

lishaduck:
🤞 I think I'm done.

package.json
```diff
@@ -4,9 +4,10 @@
   "description": "Run elm-review from Node.js",
   "type": "commonjs",
   "engines": {
-    "node": ">=14.17"
+    "node": ">=14.18"
   },
-  "packageManager": "npm@10.7.0",
+  "enginesStrict": true,
+  "packageManager": "npm@8.19.4",
```
jfmengels: Why was the `packageManager` downgraded? Is it not only used for our internal purposes, where it's set to v16?

lib/remote-template.js
```diff
@@ -107,43 +107,45 @@ async function downloadTemplateElmJson(template, commit) {
...
-
-        throw new ErrorMessage.CustomError(
-          // prettier-ignore
-          'TEMPLATE ELM.JSON NOT FOUND',
-          // prettier-ignore
-          `I found the ${chalk.yellow(repoName)} repository on GitHub,
+  let response;
+
+  try {
+    response = await got(
```
jfmengels, line 110: This is one of the cases where I think it's easier to understand the flow of values when using `.then()` rather than async/await, mostly the fact that in the former there is no `let reponse;` pending without a value, and needed to understand/check that there is no code path where `response`'s value stays `undefined`.

`.catch()` being able to handle the errors until a certain point without having to split the code with a dangling `response` is part of the reason.

I think I'd rather extract functions to make it shorter if needed, rather than this change.

```js
return got(...)
  .catch(handleError)
  .then(handleResponse);

function handleResponse(response) {
  // ...
}
```

What do you think?

lib/suppressed-errors.js
```diff
@@ -52,7 +52,9 @@ async function write(options, suppressionFiles) {
 
       const previousContents = await FS.readFile(filePath).catch(() => '');
       if (previousContents !== newContents) {
-        return FS.writeFile(filePath, newContents).catch((error) => {
+        try {
+          return await FS.writeFile(filePath, newContents);
```
jfmengels, line 56: What is the value of adding `await` after a `return`? :thinking:

lib/watch.js
```diff
@@ -1,4 +1,5 @@
 const path = require('node:path');
+// TODO [engine:node@>=19.1] Use `fs.watch` (or `@parcel/watcher`).
```
jfmengels, line 2: To be evaluated when we upgrade, but I'm not sure we will. There seem quite a few caveats for the native watcher: https://nodejs.org/docs/latest/api/fs.html#fswatchfilename-options-listener

new-package/elm-review-package-tests/helpers/find-configurations.js
```diff
@@ -10,11 +10,17 @@ module.exports = {
...
-      nodir: true
-    })
-    .map(path.dirname);
+  return (
+    glob
+      .sync(`${root}/preview*/**/elm.json`, {
+        ignore: ['**/elm-stuff/**'],
+        nodir: true
+      })
+      // eslint-disable-next-line @typescript-eslint/unbound-method -- I don't know how to fix this.
```
jfmengels, line 23: This will make it into the users' project. I'd rather have the rule disabled for this file (by defining an exception in the ESLint config, not a different disable comment at the start of the file) than to have a disable comment.

lib/suppressed-errors.js
jfmengels: Okay, the rule description gives some arguments in favor: https://typescript-eslint.io/rules/return-await/
Alright, let's try it for the sake of better stack traces :+1:

.eslintrc.js
```diff
@@ -88,7 +99,15 @@ module.exports = {
     'security/detect-object-injection': 'off',
     'security/detect-non-literal-fs-filename': 'off',
     'security/detect-non-literal-require': 'off',
-    'security/detect-unsafe-regex': 'off' // TODO: Add `eslint-plugin-regexp` and fix these issues.
+    'security/detect-unsafe-regex': 'off', // TODO: Add `eslint-plugin-regexp` and fix these issues.
+
+    // TODO: Discuss whether to enable these rules.
+    'unicorn/no-null': 'off',
+    'unicorn/prevent-abbreviations': 'off',
```
jfmengels, line 106: I'd say it's fine to use `null`, especially explicitly :man_shrugging: 

I also don't feel a need to remove the abbreviations right now :man_shrugging:

lib/autofix.js
```diff
@@ -186,26 +192,26 @@ function formatFileContent(options, file, filePath) {
   if (result.error) {
     if (result.error.code === 'ENOENT') {
       if (hasElmFormatPathFlag) {
-        throw new ErrorMessage.CustomError(
-          'ELM-FORMAT NOT FOUND',
-          // prettier-ignore
-          `I could not find the executable for ${chalk.magentaBright('elm-format')} at the location you specified:
+        throw hasElmFormatPathFlag
+          ? new ErrorMessage.CustomError(
```
jfmengels, line 196: I have to admit this was a bit jarring :sweat_smile: I hadn't seen `throw conditon ? a : b` before, and it wasn't clear that it didn't just do something like `throw hasElmFormatPathFlag ; <something else>`

Maybe simpler to have something like:

```js
const details = hasElmFormatPathFlag
  ? `I could not find the executable for ${chalk.magentaBright('elm-format')} at the location you specified: ...`
   : `the other message`;

throw new ErrorMessage.CustomError(..., details);
```

What do you think?

lib/elm-files.js
```diff
@@ -20,24 +20,6 @@ const Cache = require('./cache');
 
 const defaultGlob = '**/*.elm$';
 
-/**
- * @template T, U
- * @param {T[]} array
- * @param {(data: T) => U[]} fn
- * @returns {U[]}
- */
-function flatMap(array, fn) {
```
jfmengels, line 33: Nice :+1:

lib/new-rule.js
```diff
@@ -189,19 +189,16 @@ async function addRule(options, elmJson, ruleName, ruleType) {
   }
 
   // Inject configuration into every preview configuration
-  glob
-    .sync(OsHelpers.makePathOsAgnostic(`${dir}/preview*/**/elm.json`), {
+  for (const configurationElmJson of glob.sync(
```
jfmengels, line 192: Let's extract `glob.sync` to a variable, this is a bit hard to read.

package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "test": "turbo run testing",
     "elm-format": "elm-format --validate ast-codec init-templates new-package parseElm review template",
-    "eslint-check": "eslint . --report-unused-disable-directives",
+    "eslint-check": "eslint . --report-unused-disable-directives --max-warnings=0",
```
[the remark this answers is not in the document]
lishaduck: I set a few, and the default configs do a few as well.
I'd mostly like to set it so it looks different in the editor. Though the ESLint integration in VSCode is acting up.

package.json
lishaduck: Node 16 comes with NPM 8. I'd bumped Node a while back and then unbumped them, but forgot to revert the npm change.

lib/remote-template.js
lishaduck: Yeah, that's fair.

lib/suppressed-errors.js
lishaduck: It's faster (you'll see people saying it's slower, but it's now better optimized), it has better stack traces, and it makes try-catch work.

lib/watch.js
lishaduck: I've heard quite a few projects use `@parcel/watcher` instead, and most of the caveats were resolved in 19.1 (I think the Windows stuff is outdated).

new-package/elm-review-package-tests/helpers/find-configurations.js
lishaduck: Fair. I'll switch it when I get a few minutes.

lib/options.js
```diff
@@ -332,8 +332,8 @@ function parseUnsuppress(args) {
 }
 
 /**
- * @param {string | string[] | *} input
```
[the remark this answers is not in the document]
lishaduck: I think that would also be a valid typing.

lib/autofix.js
lishaduck: Oh, I'd meant to change this. Thanks for catching that!

lib/new-rule.js
lishaduck: `glob.sync`, or the whole expression?

package.json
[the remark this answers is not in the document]
jfmengels: Alright, that's fine by me, seeing it as different in the editor is the one place where I agree it can make sense to have rules marked as errors or warnings :smile:

lib/options.js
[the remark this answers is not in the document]
jfmengels: I'd say so (but can happen in a later PR if you prefer), because I'm pretty sure that at the moment the input is sometimes `undefined`, so it seems that TS doesn't catch that (probably because they're `any`?)

jfmengels:
Thank you for all the changes @lishaduck. I hope it will help make the project more maintainable :slightly_smiling_face: 

I'm happy to merge as-is (or rather "rebase fast-forward" the branch manually outside of GitHub, to avoid you needing to rebase the other PRs) and delay the remaining feedback for another PR, would you prefer that?

lib/options.js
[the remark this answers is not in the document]
lishaduck: Yeah, that would make sense. I can fix it.

lishaduck:
> Thank you for all the changes @lishaduck. I hope it will help make the project more maintainable ☺️
> 
> I'm happy to merge as-is (or rather "rebase fast-forward" the branch manually outside of GitHub, to avoid you needing to rebase the other PRs) and delay the remaining feedback for another PR, would you prefer that?

Eh, I don't mind rebasing, but if you'd prefer that I don't really care.

jfmengels:
I'll let you make the remaining changes then :+1:

PR #174, opened 2024-06-25
Help refactor

lishaduck:
Reland the remaining changes from #126.

Depends on #168.
Here's the diff as GH doesn't play well with stacks from forks: https://github.com/lishaduck/node-elm-review/compare/eslints...help-refactor


lishaduck:
This works, it makes more sense logically, etc. However, it no longer lazily calls `unknownCheck()`. I'm starting to wonder how worthwhile this is after all.

lishaduck:
Ok, I decided to approach this differently then #126 did. It doesn't always return true, it can also never return and exit instead. I've update the types to reflect this.

jfmengels (approved):
LGTM :+1:

PR #179, opened 2024-06-25
More Never

lishaduck:
Explicitly state "`NoReturn`s"

Depends on #174.
Here's the diff as GH doesn't play well with stacks from forks: <https://github.com/lishaduck/node-elm-review/compare/help-refactor...never>


lishaduck:
I bumped into typescript-eslint/typescript-eslint#3455.

bin/elm-review
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env node
 
+// TODO: With ESM, this should use TLA.
```
jfmengels, line 3: What is TLA? :thinking:

lib/autofix.js
```diff
@@ -46,7 +46,7 @@ function subscribe(options, app, elmVersion) {
  * @param {Options} options
  * @param {App} app
  * @param {ElmVersion} elmVersion
- * @returns {(data: AutofixRequest) => Promise<void>}
+ * @returns {(data: AutofixRequest) => Promise<void | never>}
```
jfmengels, line 49: I'm curious, does this help TS figure out more things?

lib/main.js
```diff
@@ -149,61 +165,92 @@ if either your review configuration or your project's dependencies change.`);
...
+      if (options.help) {
+        return Help.prepareOffline();
+      }
+
+      return await prepareOffline();
     }
 
-    return prepareOffline();
-  }
+    default: {
```
jfmengels, line 234: Should this not be `undefined:`, so that we get exhaustiveness errors from TS when we add new subcommands?

lib/main.js
```diff
@@ -149,61 +165,92 @@ if either your review configuration or your project's dependencies change.`);
...
+        return Help.newPackage();
+      }
 
-  if (options.subcommand === 'new-package') {
-    if (options.help) {
-      return Help.newPackage();
+      try {
+        return await NewPackage.create(options);
+      } catch (error) {
+        errorHandler(error);
```
jfmengels, line 210: Remind me: Does this `await + try catch` improve stack traces? Because otherwise I have to admit I don't see this as an improvement. It's more verbose and less Elm-like (in the sense of chaining `Task`s), which has its benefits too.

Also, shouldn't there be a `return` on or after `errorHandler`? Otherwise, I think we'll run the code after the `switch` statement when an error occurs.

bin/elm-review
lishaduck: Top Level Await

lib/autofix.js
lishaduck: To be explicit to developers. It also helps TS detect unused code, but not much.

lib/main.js
lishaduck: Hmmm... can you do that? I'll check in a few minutes.

lib/main.js
lishaduck: It should, is my understanding. I could be misremembering this though.

ErrorHandler returns never, so it's not necessary.
If you'd prefer it reverted, that's fine as well. It is more verbose.

jfmengels: I think I'd prefer it reverted :shrug:

jfmengels:
Thank you for your work on this @lishaduck!

PR #180, opened 2024-06-26
Add types to `report.js`

lishaduck:
Depends on #177.
Here's the diff as GH doesn't play well with stacks from forks: <https://github.com/lishaduck/node-elm-review/compare/app-wrapper-types...report-types>


lishaduck:
I hit rate limits.

PR #132, opened 2023-06-19
Add types to state.js

marc136:
Relates to #125 and adds types to lib/state.js

The changes actually depend on #125 which adds the definitions to `lib/options.js` . Without it, the type checker will report errors about the options file.  
For development, I added the line `// @ts-nocheck part https://github.com/jfmengels/node-elm-review/pull/126` to the top of `lib/options.js` to ignore it.

I wasn't sure if opening the PR makes sense, but I already forgot most of what I was doing there and that won't get any better :smile:

PR #181, opened 2024-06-26
Add types to `extra-files.js`

lishaduck:
Depends on #175.
Here's the diff as GH doesn't play well with stacks from forks: <https://github.com/lishaduck/node-elm-review/compare/bump-turbo...extra-files-types>

### DIFF
--- a/lib/extra-files.js
+++ b/lib/extra-files.js
@@ -16,9 +16,11 @@ const OsHelpers = require('./os-helpers');
  */
 async function collect(requests) {
   const {globby} = await import('globby');
-  const files = await Promise.all(
+  const files2D = await Promise.all(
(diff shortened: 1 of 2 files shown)


jfmengels:
I misdid the order of my Git operations, so it looks closed but this was merged :smile: 

Thanks for all the PRs @lishaduck!

jfmengels:
Do you know if the remaining open PRs are still relevant or if I should close them?

lishaduck:
> Do you know if the remaining open PRs are still relevant or if I should close them?

As in #132? I don't think it's still relevant. The Dependabot ones are probably fine if they're rebased. #147 could be closed, but until I've landed a replacement, I figure we might as well keep it open.

PR #182, opened 2024-06-26
Improve `ElmJson` types

lishaduck:
Depends on #175.
Here's the diff as GH doesn't play well with stacks from forks: <https://github.com/lishaduck/node-elm-review/compare/bump-turbo...better-elm-json-types>


lishaduck:
Ooops. I accidentally used ESM syntax. I'll fix that when I get a minute.

lishaduck:
I have another PR, but it'll need this to get rebased.

PR #184, opened 2024-06-28
Minor changelog change

lishaduck:
markdownlint caught an unused link definition.
I also added a piece of punctuation and two "to"s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


jfmengels:
Nice!

The missing "to" was in the form of "Hey, thank you @lishaduck for doing this" rather than "This was done by @lishaduck", but changing it is fine to me, maybe it will be less confusing :smile:

lishaduck:
> Nice!
> 
> The missing "to" was in the form of "Hey, thank you @lishaduck for doing this" rather than "This was done by @lishaduck", but changing it is fine to me, maybe it will be less confusing 😄

Oh, that makes sense. :)

Issue #173, opened 2024-06-25
GHA Improvements


lishaduck:
Some improvments:

- [ ] Run GHA in a matrix to catch incompatibility issues faster.
- [ ] Make some GHA security fixes.
  - [ ] Pin dependencies.
  - [ ] Add dependabot.
  - [ ] Restrict permissions.
- [ ] Update GHA dependencies.
- [ ] Add NPM Provenance.

lishaduck:
@jfmengels, I just figured out why we hit rate-limits so much: we run CI on push to any branch & on PRs, so it gets run twice.

PR #186, opened 2024-07-03
GHA Updates

lishaduck:
Improving security, etc.

_**Related:** #173_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,9 @@ defaults:
   run:
     shell: bash
 
+permissions:
+  contents: read
+
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   test:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,30 +16,30 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-    - uses: actions/checkout@v4
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
 
     - name: Setup Node.js environment
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
       with:
         node-version: 16.x
         cache: 'npm'
 
     - name: Cache ~/.elm
-      uses: actions/cache@v4
+      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
       with:
         path: ~/.elm
         key: elm-${{ runner.os }}-${{ hashFiles('**/elm.json', 'elm-tooling.json') }}
 
     - name: Cache node_modules
       id: cache-node_modules
-      uses: actions/cache@v4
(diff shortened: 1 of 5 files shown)


new-package/github/workflows/test.yml
```diff
@@ -64,23 +67,26 @@ jobs:
   publish:
     needs: [test] # make sure all your other jobs succeed before trying to publish
 
+    permissions:
+      contents: write
+
```
lishaduck, line 72: I read through the `dillonkearns/elm-publish-action` source code, and I think this is all it needs.

lishaduck:
This doesn't resolve all of #173, but the other updates were more involved, while this is self-contained (and I'm on vacation).

.github/workflows/test.yml
```diff
@@ -12,30 +12,30 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-    - uses: actions/checkout@v4
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
```
jfmengels, line 19: Is `@4` not enough? Do you think it risks breaking in the future?

new-package/github/workflows/test.yml
jfmengels: Alright. Next time I try to publish a package I'll try to backport this change there and test it out.

.github/workflows/test.yml
lishaduck: I guess for generated files, it might be reasonable not to pin them, but for us, it's more secure not to.
Pinning to a commit means that a malicious actor can't change the tag, as a commit is immutable, but not a tag. GitHub also doesn't have the greatest track record with both bugs and SemVer, and has happily broken the latest versions of actions for weeks at a time.
Dependabot should bump them, so missing patches shouldn't be a major issue.

I explained this a few weeks ago in a similar PR, so if my wording doesn't make sense (I personally doubt I said this in a way that makes sense), try this explanation instead: <https://github.com/material-extensions/vscode-material-icon-theme/pull/2365#discussion_r1639112514>.

jfmengels: What do you think about adding pinning these versions in this file and in `new-package` generated files, and adding a comment to explain the reason and how to upgrade (how/where to find the new commit for instance). I think that way both this project and users of the package benefit from the improvements here.

jfmengels:
Happy to merge this as is :+1: 
Do you mind resolving the conflict?

.github/workflows/test.yml
lishaduck: Yeah. That was a lot more noise than I'd expected.

jfmengels: > Yeah. That was a lot more noise than I'd expected.

Same here :sweat_smile: 
The GitHub rate limit has exploded :joy: 

> With elm packages though, it should just notify that there're new majors.

Hmm, I don't even know if it should. For a package author, as long as they can run tools and tests (locally and in CI) and publish (through CI), they don't need to have the latest version of anything :man_shrugging:

lishaduck: Yeah, but it would block the application from upgrading other review rules.
Say `elm-review-simplify` needs `elm-review` 2 or 3, but only declares support for 2, and `elm-review-unused` needs `elm-review` 3, then one couldn't get the improvements to `elm-review-unused`.

lishaduck: Another thing, I think I forgot to set it to weekly, I usually have it set to daily, but weekly is a lot more manageable in my experience.

PR #187, opened 2024-07-06
TS Check, Lint, and Format `new-package/` and `vendor/`

lishaduck:
Part 1 of 2.

_**Relates**: #125_

### DIFF
--- a/lib/new-package.js
+++ b/lib/new-package.js
@@ -310,6 +310,14 @@ ElmjutsuDumMyM0DuL3.elm
     }
   );
 
+  const checkPreviewsCompile = path.join(
+    packageTests,
+    'check-previews-compile.js'
+  );
+  const data = await FS.readFile(checkPreviewsCompile);
+  const result = data.replace(/\/\/ @ts-ignore - Generated file.\n/, '');
+  await FS.writeFile(checkPreviewsCompile, result, 'utf8');
+
   Spinner.succeed();
 }
 
--- a/new-package/elm-review-package-tests/check-previews-compile.js
+++ b/new-package/elm-review-package-tests/check-previews-compile.js
@@ -4,6 +4,7 @@ const path = require('path');
 const Ansi = require('./helpers/ansi');
 const {execSync} = require('child_process');
 const {findPreviewConfigurations} = require('./helpers/find-configurations');
+// @ts-ignore - Generated file.
 const packageDependencies = require('../elm.json').dependencies;
 
 const root = path.dirname(__dirname);
--- a/new-package/elm-review-package-tests/check-previews-compile.js
+++ b/new-package/elm-review-package-tests/check-previews-compile.js
@@ -1,8 +1,8 @@
 #!/usr/bin/env node
 
-const path = require('path');
+const path = require('node:path');
 const Ansi = require('./helpers/ansi');
-const {execSync} = require('child_process');
+const {execSync} = require('node:child_process');
 const {findPreviewConfigurations} = require('./helpers/find-configurations');
 // @ts-ignore - Generated file.
 const packageDependencies = require('../elm.json').dependencies;
@@ -10,8 +10,9 @@ const packageDependencies = require('../elm.json').dependencies;
 const root = path.dirname(__dirname);
 
 // Find all elm.json files
-
-findPreviewConfigurations().forEach(checkThatExampleCompiles);
+for (const example of findPreviewConfigurations()) {
+  checkThatExampleCompiles(example);
+}
 
 function checkThatExampleCompiles(exampleConfiguration) {
   const exampleConfigurationElmJson = require(`${exampleConfiguration}/elm.json`);
@@ -49,7 +50,6 @@ and make the necessary changes to make it compile.`
       }
 
       success(exampleConfiguration);
-      return;
     } catch {
       console.log(
         `An error occurred while trying to check whether the ${Ansi.yellow(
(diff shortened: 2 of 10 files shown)


lishaduck:
FYI, test failures are outdated snapshots, not #189.

new-package/elm-review-package-tests/check-previews-compile.js
```diff
@@ -82,12 +82,13 @@ function checkDepsAreCompatible(exampleConfiguration, previewDependencies) {
...
     delete previewDependencies[depName];
-  });
+  }
 
   const remainingKeys = Object.keys(previewDependencies);
-  if (remainingKeys.length !== 0) {
+  if (remainingKeys.length > 0) {
     console.error(
-      `There are extraneous dependencies in the ${exampleConfiguration}/ configuration: ${remainingKeys}`
+      // prettier-ignore
```
jfmengels, line 90: I have to admit I'm not fond in having disable checks (for prettier or for TS) in files that package authors will then own.
Maybe in this case, we can create a new variable `const keys = remainingKeys.join(', ');` above `console.error(...)`?
That should get rid of the need for the ignore.

new-package/elm-review-package-tests/check-previews-compile.js
```diff
@@ -1,16 +1,18 @@
 #!/usr/bin/env node
 
-const path = require('path');
+const path = require('node:path');
 const Ansi = require('./helpers/ansi');
-const {execSync} = require('child_process');
+const {execSync} = require('node:child_process');
 const {findPreviewConfigurations} = require('./helpers/find-configurations');
+// @ts-ignore - Generated file.
```
lishaduck, line 7: Here's another ignore comment. I figure we can call `.replace()` on the file and delete this line.

lishaduck: Did it!

lib/new-package.js
```diff
@@ -310,6 +310,14 @@ ElmjutsuDumMyM0DuL3.elm
     }
   );
 
+  const checkPreviewsCompile = path.join(
+    packageTests,
+    'check-previews-compile.js'
+  );
+  const data = await FS.readFile(checkPreviewsCompile);
+  const result = data.replace(/\/\/ @ts-ignore - Generated file.\n/, '');
+  await FS.writeFile(checkPreviewsCompile, result, 'utf8');
```
lishaduck, line 319: I'm rather proud of myself: I wrote some JS that didn't blow up in my face 🤣

new-package/elm-review-package-tests/check-previews-compile.js
jfmengels: Not optimal in terms of performance, but I guess it works :sweat_smile:

new-package/elm-review-package-tests/check-previews-compile.js
jfmengels: You marked this as resolved, but the comment is still here. Did you apply it on a different branch maybe?

new-package/elm-review-package-tests/check-previews-compile.js
lishaduck: I figure that if we go the route of hardcoding paths for #147, then this wouldn't be so bad 🤷‍♂️
Then again, that's not a performance optimal route...

jfmengels: Yes, not an issue in terms of performance. But I'm not sure I get what the relation with #147 is, sorry

lishaduck: I thought you meant that we were copying the files, then editing them, but with #147, we'll move them one by one, so we could cut out the duplication, but at the expense of total speed (as a native cp is certainly faster than our kinda-pure-js one would be.)

jfmengels: That is what I meant. Even with #147 and specifying paths we still would do this copy+edit+write process for this file, no?

Actually, can't we add a dummy `new-package/elm.json` file to get rid of the TS error entirely?

lishaduck: Yeah, I guess we're just talking past each other.
We could. Would our actual one overwrite it?

new-package/elm-review-package-tests/check-previews-compile.js
jfmengels: Let's resolve this later, let's not block all the other PRs because of this.

jfmengels:
I'll merge this PR. We can always improve things later, I don't want you to fall into Git branch hell :sweat_smile: 
And thanks again for all this work, it's incredible! :heart:

lishaduck:
> I'll merge this PR. We can always improve things later, I don't want you to fall into Git branch hell 😅 And thanks again for all this work, it's incredible! ❤️

It's a bit too late for that 🤣:
Screenshot 2024-07-08 at 5 10 26 PM

I mean, that's what summer break is for, eh? Finding projects? Enjoying not having to tune out teachers while you edit this and that? 😉

new-package/elm-review-package-tests/check-previews-compile.js
jfmengels: I think that if we get in #147 then we can skip copying the dummy `elm.json`, so maybe it should happen in tandem with #147 (which btw, feel free to take over, I don't remember the state of it).

jfmengels:
Right, but if I waited a few more days, who knows how much worse it would have been :smile:

lishaduck:
> Right, but if I waited a few more days, who knows how much worse it would have been 😄

True, true... YIKES!

new-package/elm-review-package-tests/check-previews-compile.js
lishaduck: Huh. Well, once it's all merged I can check again.

PR #188, opened 2024-07-06
Enable `noImplicitAny` on `node-elm-compiler.js`

lishaduck:
Part 2 of 2.

_**Depends on:** #187_
_**Relates**: #125_
Here's the diff as GH doesn't play well with stacks from forks: <https://github.com/lishaduck/node-elm-review/compare/ts-check...node-elm-compiler-no-implict-any>

### DIFF
--- a/vendor/node-elm-compiler.js
+++ b/vendor/node-elm-compiler.js
@@ -1,10 +1,10 @@
 'use strict';
 
 /**
+ * @import {ChildProcess} from 'node:child_process'
  * @import {CompileOptions, ProcessOptions, Sources, Spawner} from './types/node-elm-compiler'
  */
 
-// @ts-ignore - I'm offline right now.
 var spawn = require('cross-spawn');
 var elmBinaryName = 'elm';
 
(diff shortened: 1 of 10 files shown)


vendor/node-elm-compiler.js
```diff
@@ -1,8 +1,14 @@
 'use strict';
 
+/**
+ * @import {CompileOptions, ProcessOptions, Sources, Spawner} from './types/node-elm-compiler'
+ */
+
+// @ts-ignore - I'm offline right now.
```
jfmengels, line 7: Is this still relevant? :sweat_smile:

jfmengels (commented):
Looks good to me :+1: 
There is that one `// @ts-ignore` that I imagine should be removed.
And I'll wait for https://github.com/jfmengels/node-elm-review/pull/187 to be merged before merging this one.

lishaduck:
I looks like the best way to resolve that ts-ignore would be to move to `execa` (which is ESM-only), but in the meantime, I've added some `@ts-expect-error`s. I'd forgotten why it wasn't already installed—there aren't types for the latest (4-year-old![^1]) major.

[^1]: As in, for such a widely depended on project, you'd expect someone to have upgraded the types by now.

PR #190, opened 2024-07-06
Bump Engines

### LINKED ISSUE #189: Node 14, 16 installs error again
Should still be fine for now, but `glob` (and the other issacs packages: `rimraf`, `lru-cache`, and `jackspeak`) just dropped 14 and 16 _in a **patch**_, and NPM warns on old Node versions again. Tests aren't catching it, and ls-engines doesn't resolve based on the ideal tree, just the lockfile. Though I think a flag should fix that, at least.

https://github.com/isaacs/node-glob/commit/8a69def3cad0de9ba26ca831065ffe448d153de3

_**CC:** @jfmengels_

lishaduck:
Resolves #189.


lishaduck:
This'll fail CI becuase there are some other changes that should go along with the upgrade, but I'm not going to put in that effort if we find a better way to deal with this.

lishaduck:
Ok, I went ahead and "suppressed" the eslint failures, so that there's a green check, i.e., this is otherwise fine if this is how we decide to do this.

lishaduck:
Might've found a workaround: getsentry/sentry-javascript#12793
I'll try it.

lishaduck:
> Might've found a workaround: [getsentry/sentry-javascript#12793](https://github.com/getsentry/sentry-javascript/pull/12793) I'll try it.

No, that would prevent people from upgrading the packages on newer node. It should be set by application developers.

package.json
```diff
@@ -114,7 +114,7 @@
   },
   "packageManager": "npm@8.19.4",
   "engines": {
-    "node": ">=16 || 14 >=14.18"
+    "node": ">=18"
```
jfmengels, line 117: Do we not have a way around this? Can't we pin the version of `glob` to one that supports the previous version?

I just saw that v10.4.4 was released and supports `"14 >=14.21 || 16 >=16.20 || 18 || 20 || >=22"`. This is again higher than what we had before, but maybe better than dropping 14 and 16?

Although all of this is starting get pretty annoying and makes me want to rip off the bandaid and drop support so that we're done with this... but also I think that's wrong, so I'm trying really hard to resist... I'm not sure what to do yet.

lishaduck: > I just saw that v10.4.4 was released and supports `"14 >=14.21 || 16 >=16.20 || 18 || 20 || >=22"`. This is again higher than what we had before, but maybe better than dropping 14 and 16?
 
Yeah. I guess we just match that.

> Although all of this is starting get pretty annoying and makes me want to rip off the bandaid and drop support so that we're done with this... but also I think that's wrong, so I'm trying really hard to resist... I'm not sure what to do yet.

I wouldn't mind, but in a patch? Ruining everyone's resolution? I'm rather unhappy about it. And it wasn't even needed. Yes, `??=` is nice, but he needed it once. And it's just syntax sugar. And just in the cli that very few even use in the first place.

lishaduck: Either way, there're some other nice changes to be had here, I'll try to break them out.

lishaduck:
Ok, this PR now does the minimal work to make us compatible with latest glob.

lishaduck:
Glob 11 just drops support again, I guess we should declare support for it as well.

lishaduck:
Ok, I guess we just have to trust ls-engines to keep us up-to-date here. We need to support Glob 11, but npm doesn't take the engine into account.

PR #191, opened 2024-07-06
Enable `noImplicitAny` on `result-cache-json.js`

lishaduck:
_**Depends on:** #188_
_**Relates**: #125_
Here's the diff as GH doesn't play well with stacks from forks: <https://github.com/lishaduck/node-elm-review/compare/node-elm-compiler-no-implict-any...result-cache-json-no-implict-any>

### DIFF
--- a/lib/result-cache-json.js
+++ b/lib/result-cache-json.js
@@ -3,7 +3,15 @@ module.exports = {
   reviver
 };
 
-function replacer(_key, value) {
+/**
+ * @import {Replacer, Reviver} from "./types/json"
+ * @import {Value, ValueStruct} from "./types/elm-internals"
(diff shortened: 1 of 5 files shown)


lib/result-cache-json.js
```diff
@@ -41,7 +49,7 @@ function replacer(_key, value) {
   }
 
   if (typeof value !== 'number' || Number.isFinite(value)) {
-    return value;
+    return value ?? null;
```
lishaduck, line 52: This was unintentional.
Well, I intended it initially, but fixed some other type errors, so this should be unnecessary.

lishaduck:
Rate limits.

PR #193, opened 2024-07-06
Improve shell scripts

lishaduck:
I'm going to run our testing scripts through [OSH](https://www.oilshell.org/cross-ref.html?tag=OSH#OSH), see what antipatterns it spots, and fix the warnings.
OSH is a fast, simple, (almost) 100% bash-compatible and POSIX-compliant shell from the Oils project. It supports some extra `shopt`s, which catch things bash and shellcheck don't/can't (respectively). It also supports some `shopt`s that make breaking changes (ysh), but I don't plan on making our tests non-compatible with bash (as that would break CI, for one reason).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,10 @@ name: CI
 
 on: [push, pull_request]
 
+defaults:
+  run:
+    shell: bash
+
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   test:
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ ElmjutsuDumMyM0DuL3.elm
 .vscode/
 
 .turbo
+.envrc
--- a/test/run.sh
+++ b/test/run.sh
@@ -215,7 +215,7 @@ $createTest "$CMD" \
 
 if [ "$1" != "record" ]
 then
-    if [ "$(diff -yq "$TMP/project to fix/src/Main.elm" "$SNAPSHOTS/project to fix/src/Main.elm")" != "" ]
+    if [ "$(diff -q "$TMP/project to fix/src/Main.elm" "$SNAPSHOTS/project to fix/src/Main.elm")" != "" ]
     then
         echo -e "Running with --fix-all-without-prompt (looking at code)"
         echo -e "\x1B[31m  ERROR\n  I found a different FIX output  for Main.elmthan expected:\x1B[0m"
@@ -223,7 +223,7 @@ then
         diff -py "$TMP/project to fix/src/Main.elm" "$SNAPSHOTS/project to fix/src/Main.elm"
         exit 1
     fi
-    if [ "$(diff -yq "$TMP/project to fix/src/Folder/Used.elm" "$SNAPSHOTS/project to fix/src/Folder/Used.elm")" != "" ]
+    if [ "$(diff -q "$TMP/project to fix/src/Folder/Used.elm" "$SNAPSHOTS/project to fix/src/Folder/Used.elm")" != "" ]
     then
         echo -e "Running with --fix-all-without-prompt (looking at code)"
         echo -e "\x1B[31m  ERROR\n  I found a different FIX output than expected for Folder/Used.elm:\x1B[0m"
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -37,3 +37,4 @@ ElmjutsuDumMyM0DuL3.elm
 .vscode/
 
 .turbo
+.envrc
```
lishaduck, line 40: I got tired of all of the rate limits.

test/run.sh
```diff
@@ -1,20 +1,20 @@
 #!/bin/bash
 
-set -e
+set -euo pipefail
```
lishaduck, line 3: `-o pipefail` fails commands if an intermediate step fails.
`-u` is equivalent to `-o nounset`, which fails a script if it references unbound variables.

test/run.sh
```diff
@@ -1,20 +1,20 @@
...
 
 CWD=$(pwd)
 CMD="elm-review --no-color"
 TMP="$CWD/temporary"
 ELM_HOME="$TMP/elm-home"
 SNAPSHOTS="$CWD/run-snapshots"
-SUBCOMMAND="$1"
+SUBCOMMAND="${1:-}"
```
lishaduck, line 10: This makes the subcommand default to none if there is none passed, for `nounset`.

test/run.sh
```diff
@@ -34,10 +34,12 @@ function runCommandAndCompareToSnapshot {
       exit 1
     fi
 
-    eval "$LOCAL_COMMAND$AUTH --FOR-TESTS $ARGS" 2>&1 \
+    (eval "$LOCAL_COMMAND$AUTH --FOR-TESTS $ARGS" || true) 2>&1 \
```
lishaduck, line 37: All of these `|| true`s tell bash that the command might fail. They should be here either way, it's a correctness issue, but pipefail enforces it.
The subshell makes it so that it's not the `true` that gets redirected to `/dev/null/`, but the whole thing.

test/run.sh
```diff
@@ -34,10 +34,12 @@ function runCommandAndCompareToSnapshot {
...
         | $REPLACE_SCRIPT \
         > "$TMP/$FILE"
-    if [ "$(diff "$TMP/$FILE" "$SNAPSHOTS/$FILE")" != "" ]
+    local diffed
+    diffed="$(diff "$TMP/$FILE" "$SNAPSHOTS/$FILE" || true)"
+    if [ "$diffed" != "" ]
```
lishaduck, line 42: This is something OSH's `strict_errexit` caught. Before, if diff failed, it would be ignored. That's not good, so it has to be extracted into a variable.
However, we do want to do our own handling of it, so we then explicitly ignore the potential failures.

test/run.sh
```diff
@@ -46,7 +48,6 @@ function runCommandAndCompareToSnapshot {
         cat "$TMP/$FILE"
         echo -e "\n    \x1B[31mHere is the difference:\x1B[0m\n"
         diff -p "$TMP/$FILE" "$SNAPSHOTS/$FILE"
-        exit 1
```
lishaduck, line 49: diff exits with exit code one, so this was dead code.

test/run.sh
```diff
@@ -153,7 +156,7 @@ rm -rf "$TMP" \
 
 mkdir -p "$TMP"
 
-if [ "$1" == "record" ]
+if [ "$SUBCOMMAND" == "record" ]
```
lishaduck, line 159: Instead of using the potentially unbound `$1`, use the bound-to-be-bound `$SUBCOMMAND`, which also reads easier.

test/run.sh
```diff
@@ -254,7 +260,7 @@ $createTest "$CMD" \
     "Fixing all errors for an entire rule should remove the suppression file" \
     "" \
     "suppressed-errors-after-fixed-errors-for-rule.txt"
-if [ -f review/suppressed/NoUnused.Dependencies.json ]; then
+if [ -f "./review/suppressed/NoUnused.Dependencies.json" ]; then
```
lishaduck, line 263: Always put paths in strings. Doesn't make a difference here, but doesn't hurt.

.github/workflows/test.yml
```diff
@@ -4,6 +4,10 @@ name: CI
 
 on: [push, pull_request]
 
+defaults:
+  run:
+    shell: bash
```
lishaduck, line 9: Run commands with pipefail.
See <https://github.com/lowlighter/libs/pull/69#issuecomment-2212241938:~:text=It%20runs%20with,faster)%20(ok%20reason)> for a longer explaination.

test/run.sh
```diff
@@ -213,31 +216,34 @@ $createTest "$CMD" \
...
     then
         echo -e "Running with --fix-all-without-prompt (looking at code)"
         echo -e "\x1B[31m  ERROR\n  I found a different FIX output than expected for Folder/Used.elm:\x1B[0m"
         echo -e "\n    \x1B[31mHere is the difference:\x1B[0m\n"
         diff -py "$TMP/project to fix/src/Folder/Used.elm" "$SNAPSHOTS/project to fix/src/Folder/Used.elm"
-        exit 1
     fi
-    if [ "$(diff -yq "$TMP/project to fix/src/Folder/Unused.elm" "$SNAPSHOTS/project to fix/src/Folder/Unused.elm")" != "" ]
+    declare diffed
+    diffed="$(diff -q "$TMP/project to fix/src/Folder/Unused.elm" "$SNAPSHOTS/project to fix/src/Folder/Unused.elm" || true)"
```
lishaduck, line 240: `-y` and `-q` are incompatible, at least on macOS.

jfmengels:
Thank you for the improvements!

PR #196, opened 2024-07-08
Fix snapshot

lishaduck:
There was a snapshot in the root dir I've puzzled over for a while. I figured it out yesterday, so here's a fix.


jfmengels:
Nice! I hadn't even noticed it :sweat_smile:

PR #215, opened 2024-07-09
Put `ls-engines` in library mode

lishaduck:
_**Split off from:** #190_

### DIFF
--- a/ast-codec/elm.json
+++ b/ast-codec/elm.json
@@ -6,24 +6,19 @@
     "elm-version": "0.19.1",
     "dependencies": {
         "direct": {
-            "MartinSStewart/elm-serialize": "1.2.0",
+            "MartinSStewart/elm-serialize": "1.3.1",
             "elm/core": "1.0.5",
             "elm/json": "1.1.3",
-            "stil4m/elm-syntax": "7.1.3"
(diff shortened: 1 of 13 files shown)


lishaduck:
This failing is good, it's catching that we now need #190.

lishaduck:
Great, more fun! Something else must've dropped node 14- in the whole glob fiasco. Welp. Whadawedonow?

jfmengels:
:man_facepalming: :face_exhaling:

lishaduck:
Oh, just saw this! npm/npm-pick-manifest#33 fixes a longstanding bug in npm, so engines-based resolution is a thing, just like it should be. Therefore, we can revert all of these engines changes, and npm won't install the newer glob/jackspeak/etc. However, if we support node 14, then we support npm 6 (and 8 and 10), so... no 😢
The next time this happens though, it should be much less of an issue 🎉

lishaduck:
I've fixed this locally, need to split the giant branch into reviewable PRs.

lishaduck:
Note: I did fix this, but I don't want to push it till #256 gets merged b/c my branches are complicated enough as-it.

jfmengels:
Superseded by https://github.com/jfmengels/node-elm-review/pull/274

PR #217, opened 2024-07-09
Standardize comments

lishaduck:
_**Split off from:** #190_

~~_**Depends on:** #216_
Here's the diff as GH, doesn't play well with stacks from forks: <https://github.com/lishaduck/node-elm-review/compare/check-eslint-with-eslint...todo-comments>~~


lib/project-json-files.js
```diff
@@ -176,7 +176,7 @@ async function readFromPackagesWebsite(
   packageVersion,
   file
 ) {
-  // TODO [engine:node@>=18]: We can use `fetch` now.
+  // TODO(@lishaduck) [engine:node@>=21]: We can use `fetch` now.
```
jfmengels, line 179: Fetch is available in 18, no?

lishaduck: I thought so, but apparently it was only experimental until 21.

PR #218, opened 2024-07-09
Enable incremental TypeScript checking

lishaduck:


jfmengels:
Nice! I wonder why this is not enabled by default 🤔

PR #219, opened 2024-07-15
Optimize >> and << in compiled JS

jfmengels:
This effectively changes Elm's core `>>` and `<<` from (roughly)

```elm
(<<) g f x =
  g (f x)

-- to

(<<) g f =
  \x ->
    g (f x)
```
In other words, the composition functions are now wrapped in `F2` instead of `F3`, which yields much better performance because it gets applied using `A2(composeL, f, g)` in practice.

### DIFF
--- a/lib/optimize-js.js
+++ b/lib/optimize-js.js
@@ -386,6 +386,30 @@ const performanceReplacements = [
       $elm$core$Set$empty,
       set);
   });`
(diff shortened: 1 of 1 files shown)


jfmengels:
Reflects the changes done in https://github.com/elm/core/pull/1145

PR #220, opened 2024-07-16
Fix colors not showing up

jfmengels:
Colors were not showing up anymore when you running a plain `elm-review`.
It seems it was broken in 69ac731 when turning the report into `undefined`, though I'm not entirely sure how that's related :thinking: 

cc @lishaduck

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -423,7 +423,7 @@ function compileElmProject(
           )
         );
         compileOptions.processOpts.stdio = ['ignore', 'ignore', 'inherit'];
-        compileOptions.report = undefined;
+        compileOptions.report = null;
         const compileProcessToGetColoredErrors = elmCompiler.compile(
           compileTargets,
           compileOptions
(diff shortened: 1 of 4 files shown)


lishaduck:
> It seems it was broken in [69ac731](https://github.com/jfmengels/node-elm-review/commit/69ac731db18ab01463d8a71530900e71b9b9abb7) when turning the report into `undefined`, though I'm not entirely sure how that's related 🤔

That commit link isn't right. Anyway, I think I recall some type conflicts, this should work as well.

lishaduck:
> That commit link isn't right.

0de6faae0d1dfe005574857099f4acdf456f5756, for reference.

> Anyway, I think I recall some type conflicts, this should work as well. Not sure why that ts-expect-error isn't needed anymore 🤷‍♂️

I think this is correct, though I still don't really know why such an innocuous change made such a difference, especially given  this project's code coverage.
And I guess that ts-expect-error was added in 0de6faae0d1dfe005574857099f4acdf456f5756, so I guess it makes sense, a bit?

lishaduck:
Oh, and Prettier's also failing in CI :)
That ternary can be moved onto the same line again.

jfmengels:
> That commit link isn't right.

My bad :facepalm:

PR #197, opened 2024-07-08
Improve Tests (fix ci, enable typesafety, better unix/posix stuff, etc)

lishaduck:
_**Depends on:** #196._
_**Split off from:** #196._

PR #236, opened 2024-08-29
Refactor the `npx` replacement

### LINKED ISSUE #224: Support for locally installed binaries
I just realized that my patch in 43942a121b77a3c12aa96f44104773f2216a6664 masks the real issue, that `elm-review` assumes that if you've used npm to install Elm, you're also running `elm-review` via `npx`. This assumption isn't true, and we should modify the actual code to get elm from `./node_modules/bin`, and also `../node_modules/bin`, etc.

Nonetheless, I'd argue that that patch is at least a band-aid, and this can be deferred for a bit as no-one has reported this in the wild as far as I know.

EDIT (for @lydell): For a "real" use case, the tests install `elm-review` globally, but `elm` locally.

lishaduck:
There wasn't really consensus on the original issue, but I wanted to write some code, so 🤷‍♂️
I improved the docs at least :)
For reference, the diff is -82 +82 lines, so it's literally a net neutral w/regard to maintainability.
It primarily does so by reusing existing code for emulating `npx` for `elm-format` instead of using another (hacky) handrolled solution I introduced in #197 (as a nice side effect, we test that the CLI works without `--compiler` in the unit tests again).

Closes #224.


jfmengels (approved):
Thanks!

PS/heads-up: leaving on vacation for 1-2 weeks. I'll likely see the PRs but I won't have a laptop.

PR #256, opened 2024-09-25
Bump deps

lishaduck:
Follow up to #255, bumps everything else.

---

I also had this commit, but npm dedupe reverted it, so I did too:

```
chore(deps): bump elm-solve-deps-wasm

Manually edited the `package-lock.json` to force `elm-solve-deps-wasm` v2.
Then, regenerated the lockfile to give `elm-test` the old version.
Workaround for rtfeldman/node-test-runner#636.
```

Ref: rtfeldman/node-test-runner#636

PR #263, opened 2024-10-15
Boost dependabot's signal-to-noise ratio significantly

lishaduck:
Group a whole bunch of elm dependency bumps across the tree. Also automate updating the test snapshots as well. Previously, I hadn't really grouped elm dependency bumps, and they still needed to be accompanied with a manual update of the test snapshots.

@jfmengels, I think this'll help a lot. Probably merge #256 first though, as otherwise this'll make dependabot recreate a bunch of prs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,6 +25,22 @@ updates:
           - 'turbo'
           - 'eslint-config-turbo'
           - 'eslint-plugin-turbo'
+    ignore:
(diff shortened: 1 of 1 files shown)


lishaduck:
CI passed? Huh. Turbo's intelligently skipping this, but thinks it passed. Back to #264!

jfmengels:
:heart:

PR #264, opened 2024-10-18
Update Turbo Config

lishaduck:
I didn't realize Turbo automatically strips the env nowadays, so tests don't pick up the token. This should significantly reduce how often we get rate-limited in tests.


lishaduck:
Note: CI is failing b/c CI is failing, it nothing to do with this PR. See #256.

turbo.json
lishaduck, line 1: I didn't format this, so prettier's complaining.
Will get to it... someday.

lishaduck:
If we don't hit rate limits, this is good to go from me 👍

jfmengels:
This looks really nice! Thank you :relaxed:

PR #274, opened 2024-10-29
Last dependency changes branch

lishaduck:
@jfmengels, 🤞

Supersedes #215, #270, #271, #272.

Let's see what CI says.

PR #280, opened 2024-10-30
Refresh lockfile, update dependabot ignores

lishaduck:
Follow up to #274.

---

Closes #277
Closes #279


lishaduck:
@jfmengels, I saw you were on, so pinging you :)
I'd like get through this batch of deps stuff so I can get back to async and ts stuff.

PR #284, opened 2024-10-30
Count file and global errors separately

### LINKED ISSUE #282: Improve error count for global errors
`I found 39 errors in 1 file.` is misleading when those are all global errors.

What about "I found X errors in Y file[s] and Z global errors"?

jfmengels:
Fixes #282


lishaduck:
I think the test is flake, given that the pull_request trigger worked fine.

jfmengels:
I'm hitting real test failures locally at least (configuration errors don't get shown), I need to do a bit more work on this.

PR #288, opened 2024-10-31
CI improvements

lishaduck:
Endlessly chasing that 1min ci time...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
       if: steps.cache-node_modules.outputs.cache-hit != 'true'
       env:
         NO_ELM_TOOLING_INSTALL: 1
-      run: npm ci
+      run: npm ci --engine-strict
 
     - name: elm-tooling install
       run: npx --no-install elm-tooling install
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,13 +50,13 @@ jobs:
         set-safe-directory: true
 
     - name: Setup Node.js environment
-      uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+      uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
       with:
         node-version: 16.x
         cache: 'npm'
 
     - name: Cache ~/.elm
-      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
       with:
         path: ~/.elm
         key: elm-${{ runner.os }}-${{ hashFiles('**/elm.json', 'elm-tooling.json') }}
@@ -65,7 +65,7 @@ jobs:
 
     - name: Cache node_modules
       id: cache-node_modules
-      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
       with:
         path: node_modules
         key: node_modules-${{ runner.os }}-${{ hashFiles('package-lock.json') }}
@@ -73,7 +73,7 @@ jobs:
           node_modules-${{ runner.os }}-
 
     - name: Cache turbo build setup
-      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
       with:
         path: .turbo
         key: ${{ runner.os }}-turbo-${{ github.sha }}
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -116,5 +116,19 @@
...
+    "runtime": [
+      {
+        "name": "node",
+        "version": "^16.20"
+      }
+    ],
+    "packageManager": [
+      {
+        "name": "npm",
+        "version": "^8.19.4"
```
jfmengels, line 130: Interesting!

package.json
```diff
@@ -113,7 +113,7 @@
     "turbo": "^2.1.2",
     "typescript": "~5.6.2"
   },
-  "packageManager": "npm@8.19.4",
+  "packageManager": "npm@8.19.4+sha512.dc700d97c8bd0ca9d403cf4fe0a12054d376f048d27830a6bc4a9bcce02ec42143cdd059ce3525f7dce09c6a4e52e9af5b996f268d8729c8ebb1cfad7f2bf51f",
```
jfmengels, line 116: Can you explain how to find the hash for this?
So that if I want to update it some point I know how to keep this security improvement.

Do you also have more information/resources on why this is a good idea?

.github/workflows/test.yml
```diff
@@ -2,7 +2,28 @@
...
+      - opened
+      - synchronize
+      - reopened
+
+  # Support merge queues.
+  merge_group:
+
+  # Run on a schedule.
+  schedule:
+    - cron: "0 7 * * 1" # Every Monday at 9 in the morning CET
```
jfmengels, line 23: I'm not sure running on a schedule is necessary. The tests are not generally flaky (except for the GitHub rate limits) so testing `main` + PRs should be good enough.

.github/workflows/test.yml
```diff
@@ -60,13 +60,17 @@ jobs:
       with:
         path: ~/.elm
         key: elm-${{ runner.os }}-${{ hashFiles('**/elm.json', 'elm-tooling.json') }}
+        restore-keys: |
+          elm-${{ runner.os }}-
```
jfmengels, line 64: I've read https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows but I still don't get how this is not dangerous.

If `elm.json` has changed, then recovering from other caches seems like a recipe for problems, though I don't understand what it recovers. What am I missing?

.github/workflows/test.yml
```diff
@@ -50,13 +50,13 @@ jobs:
         set-safe-directory: true
 
     - name: Setup Node.js environment
-      uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+      uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
```
jfmengels, line 53: Maybe dependabot is not updating them because of the hash? Or because of the comment. Not sure either :man_shrugging:

package.json
lishaduck: I wrote it manually originally, but remembered a few days ago (in a personal project) that there's a cli, so getting the hash is as simple as `corepack use npm@x.x.x`, and it'll add the hash in automatically.

[Corepack then does an integrity check when downloading](https://github.com/nodejs/corepack?tab=readme-ov-file#when-authoring-packages:~:text=The%20hash%20is%20optional%20but%20strongly%20recommended%20as%20a%20security%20practice.)

.github/workflows/test.yml
lishaduck: Say I've got an `elm.json` with `elm-review@2.14.1` and `elm-syntax@8.3.4`. I bump `elm-syntax`, so the old cache is invalidated, but it'll still grab the last cache from the same branch, then `main` (so you can't get a cache from a fork and get a supply-chain attack), so `elm make` doesn't have to redownload the unchanged `elm-review`. Then, it'll upload the new cache (with the new `elm-syntax`) with a new hash.
It has the potential to create larger caches, but if it becomes a problem, you can just clear the cache.

.github/workflows/test.yml
lishaduck: Dependabot is smart enough for it to work on my other projects. If the comment is wrong, it'll also override it (rather than ignoring it or not updating), which is nice.

package.json
lishaduck: Did realize that there's some bugs and uncompleted functionality surrounding this, so I might revert this for now, unfortunately (npm/cli#7888 & commentary in npm/cli#7766).

.github/workflows/test.yml
jfmengels: > If there aren't flakes, I'm happy to get rid of it, but it might be useful as a reminder to update elm-syntax in tests.

Won't we get dependabot PRs for this anyway?

lishaduck: Oh, yeah, I forgot I did that 🤦‍♂️

jfmengels:
Nice!

Btw heads up, I delete all GitHub Actions cache, because some other PRs started failing for very odd reasons, so I'm suspecting that some PRs CI tests affect each other. I hope I'm wrong because that would suck. So the cache cleanup is not related to this PR, and you might get slower builds than expected if you were hoping to see them be real quick thanks to the cache.

lishaduck:
> Btw heads up, I delete all GitHub Actions cache, because some other PRs started failing for very odd reasons, so I'm suspecting that some PRs CI tests affect each other.

Hm. Cache pollution should only be possible if something snuck into main, so 🤷‍♂️

> I hope I'm wrong because that would suck. So the cache cleanup is not related to this PR, and you might get slower builds than expected if you were hoping to see them be real quick thanks to the cache.

That was (mostly) a joke. Thanks for letting me know though.

PR #253, opened 2024-09-19
Big PR [DO NOT MERGE]

lishaduck:
XC has taken a big drain out of my time and energy, and I'm primarily posting this for posterity.
Someday, I'll rebase this and make it reviewable, for now, I just want to note some improvements that could be made by someone, hopefully me?

P.S., I know no stress 🙃  That's _why_ I'm posting this, to put it somewhere I can know it's backed up and be accessed as needed.

---

Children:
- #255
- #256
- #288
- #290
- #291
- #292
- #297
- #298

PR #290, opened 2024-10-31
Sort requires

lishaduck:
Extracted from #253.

No, there's not an eslint rule to enforce this, sadly. I spent a few hours some night in late August when I was procrastinating and didn't want to get my homework done 😉


jfmengels:
There are a few good things mixed in with this (and I don't want to block your other work that may depend on this), so I've merged it, but this is also not super necessary either. In the sense that if you would have asked me whether we should do something about import sorting, I would likely have told you no. Plus, it was kind of sorted to my own heuristic, now it's sorted to a heuristic I don't know. No worries, but I think there are more interesting things to be worked on in the future :man_shrugging:

lishaduck:
> super necessary either. In the sense that if you would have asked me whether we should do something about import sorting, I would likely have told you no.

Fair enough.

> Plus, it was kind of sorted to my own heuristic, now it's sorted to a heuristic I don't know.

It looked alphabetical to me, so I thought I'd messed it up.

> No worries, but I think there are more interesting things to be worked on in the future 🤷‍♂️

Agreed.

PR #291, opened 2024-10-31
Tweak docs

lishaduck:
Extracted from #253.

Conflicts with #290, please merge that first.


README.md
```diff
@@ -6,7 +6,7 @@ Run [`elm-review`] from Node.js.
 
 ## Installation
 
-```bash
```
jfmengels, line 9: What's the difference in practice? Asking because I always do `bash`.

lib/options.js
```diff
@@ -1,3 +1,14 @@
+/**
+ * This file contains all the options from which elm-review derives how it will behave.
+ *
+ * Some of these options are not documented in `--help`.
+ * DO NOT DEPEND ON THEM!
+ * They might get removed at any point.
+ * They are not part of the user-facing public API and may be remove in a SemVer patch.
```
jfmengels, line 7:
```suggestion
 * They are not part of the user-facing public API and may be removed in a SemVer patch.
```

README.md
lishaduck: Purely semantic, as far as I can tell. I wrote this two months ago, so 🤷‍♂️
I don't think it makes a difference to syntax highlighting (on gh at least), etc, but I use zsh, Simon uses fish, etc.
I just always use `sh` for shell stuff unless I need a specific shell.

lishaduck:
Oh, tests are failing b/c of new elm-syntax. I'll send in another PR 🤣

PR #292, opened 2024-10-31
Concurrency

lishaduck:
Theoretically contains some zalgo, improves some stacktraces. Practically reduces cognitive overhead and consistency with the ecosystem.

Here's the diff as GH doesn't play well with stacks from forks: https://github.com/lishaduck/node-elm-review/compare/sort-imports...async

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -64,6 +64,7 @@ module.exports = {
     'promise/prefer-await-to-then': 'off',
     'no-return-await': 'off',
     '@typescript-eslint/return-await': ['error', 'always'],
+    '@typescript-eslint/no-confusing-void-expression': 'error',
     'unicorn/no-array-callback-reference': 'off',
     'unicorn/no-array-reduce': 'off',
     'unicorn/prefer-module': 'off',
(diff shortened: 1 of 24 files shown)


jfmengels:
> improve stacktraces by requiring async code to be async

Question out of curiosity: Can you explain how/why stack traces are improved by this change? I'd love an example of before/after so that I can have a clearer understanding?

lishaduck:
> Can you explain how/why stack traces are improved by this change? I'd love an example of before/after so that I can have a clearer understanding?

If you await a promise, it gets added to the stack, if you return it, it doesn't. The `return-await` rule doesn't trigger in sync code, so by making them async, you fix that. Essentially, all of the footguns from `return-await`.
By requiring functions that return a promise to be async, you a) prevent zalgo (sync rejection/fulfillment from a promise-returning function) and b) encourage awaiting promises over returning them.

TSESlint docs:
- https://typescript-eslint.io/rules/promise-function-async/
- https://typescript-eslint.io/rules/return-await/

jfmengels:
Thank your for the great explanation :blush:

PR #293, opened 2024-11-01
New elm-syntax

lishaduck:
Bumps elm-syntax

---

For reference, this invocation's worked for me the last few times I needed to do this: `fd --ignore --type f elm.json . --exec npx elm-json upgrade --yes`.
The only issue is that, for some reason, most (but not all) of the snapshots strip the trailing newline from the `elm.json` file.

### DIFF
--- a/ast-codec/elm.json
+++ b/ast-codec/elm.json
@@ -8,7 +8,7 @@
         "direct": {
             "elm/core": "1.0.5",
             "elm/json": "1.1.3",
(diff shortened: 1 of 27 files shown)


jfmengels:
> this invocation's worked for me the last few times I needed to do this: fd --ignore --type f elm.json . --exec npx elm-json upgrade --yes.

Interesting. I tend to run `GITHUB_AUTH=... npm run test-run-record`, but that's on the slow side for sure.

PR #295, opened 2024-11-01
Break unparsable elm.json more

jfmengels:
Supersedes #294

This will prevent dependabot from trying to update it (which I imagine it does because support for comments is improving?)

### DIFF
--- a/test/config-unparsable-elmjson/elm.json
+++ b/test/config-unparsable-elmjson/elm.json
@@ -4,7 +4,7 @@
         "src"
     ],
     "elm-version": "0.19.1",
-    "dependencies": {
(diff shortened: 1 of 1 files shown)


lishaduck:
> I imagine it does because support for comments is improving?

I assumed that too, but I just looked through it, it like is hasn't changed in over a year (beyond rubcop and sorbet fixes, and it used to shell out to elm if I understood correctly, though I didn't look too closely)

lishaduck:
Oh, I guess it passed b/c it's not considered part of cache key. I'll check.

lishaduck:
Nope, freak accident, I gues.

PR #297, opened 2024-11-03
Replace `rimraf` with native `rm`

lishaduck:
We should check if this regresses anything on Windows before merging.

### DIFF
--- a/lib/fs-wrapper.js
+++ b/lib/fs-wrapper.js
@@ -80,8 +80,8 @@ function mkdirpSync(dir) {
  * @param {fs.PathLike} file
  * @param {fs.RmOptions} [options={recursive: true}]
  */
-function remove(file, options) {
-  return fsp.rm(file, {recursive: true, ...options});
+async function remove(file, options) {
+  await fsp.rm(file, {recursive: true, ...options});
 }
 
 module.exports = {
--- a/lib/fs-wrapper.js
+++ b/lib/fs-wrapper.js
@@ -4,7 +4,6 @@
  */
 const fs = require('graceful-fs');
 const fsp = fs.promises;
-const {rimraf} = require('rimraf');
 
 /**
  * Read a JSON file.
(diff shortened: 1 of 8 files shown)


lishaduck:
Also, the test failure looks legit.

lib/fs-wrapper.js
```diff
@@ -90,7 +89,7 @@ module.exports = {
 
   readdir: fsp.readdir,
 
-  remove: rimraf,
+  remove: fsp.rm,
```
jfmengels, line 92: I imagine we need to turn on the `recursive` option on by default?
https://nodejs.org/api/fs.html#fspromisesrmdirpath-options

lishaduck: Yeah, probably.
It's deprecated, not sure why.

lishaduck: Oh, rmdir recursive is deprecated in favor of rm recursive.

lib/fs-wrapper.js
```diff
@@ -78,10 +78,13 @@ function mkdirpSync(dir) {
...
  * @param {fs.PathLike} file
- * @param {fs.RmOptions} [options={recursive: true}]
  */
-async function remove(file, options) {
-  await fsp.rm(file, {recursive: true, ...options});
+async function remove(file) {
+  await fsp.rm(
+    file,
+    // Rimraf compatibility:
+    {recursive: true, force: true, maxRetries: 10}
```
jfmengels, line 86: `rimraf` retries 10 times? Wow, that's surprising. I'm not sure I understand why it does that. I guess it's fine to keep...? There's not many downsides to it, but it's surprising that there is a need to retry this kind of operation.

jfmengels:
Thanks! One less dependency!

lib/fs-wrapper.js
lishaduck: rimraf retries 10 times _with an exponential backoff_! We just use the default _linear backoff_, otherwise we might as well just be using rimraf.

PR #298, opened 2024-11-03
Misc

lishaduck:
Last few tiny remaining commits from #253.

### DIFF
--- a/lib/result-cache-json.js
+++ b/lib/result-cache-json.js
@@ -12,7 +12,7 @@ function replacer(_key, value_) {
   }
 
   if (typeof value === 'object') {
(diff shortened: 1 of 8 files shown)


lib/result-cache-json.js
```diff
@@ -12,7 +12,7 @@ function replacer(_key, value_) {
   }
 
   if (typeof value === 'object') {
-    // TODO(@jfmengels): Also support serializing `Dict`s.
+    // // TODO(@jfmengels): Also support serializing `Dict`s.
```
jfmengels, line 15: Stray double comment addition?

PR #299, opened 2024-11-04
Update tooling docs for new executable path autoresolution

lishaduck:
Follow up to #49, #236.


lishaduck:
Let's see how fast CI is nowadays.

jfmengels:
Thanks!

PR #300, opened 2024-11-04
Parallelize engines check in CI

lishaduck:


jfmengels:
Let's go even faster! :D

PR #301, opened 2024-11-04
More TS

### LINKED ISSUE #125: Add types to JS files

lishaduck:
Slowly, steadily, yet always onward, pushing forward, toward the goal.

In other words, more types. ~~Or rather, better types, as this was mostly fixing mistakes.~~ (Nope, I ended up just finishing #125)

Resolves #125.


lib/remote-template.js
```diff
@@ -108,6 +119,10 @@ async function getRemoteElmJson(
   return TemplateDependencies.update(options, elmJson);
 }
 
+/**
+ * @param {Template} template
+ * @param {string} commit
+ */
```
jfmengels, line 125: Should this not a return type annotation? (Same of other functions in this file)?

lib/build.js
```diff
@@ -352,13 +357,22 @@ function updateSourceDirectories(options, userSrc, elmJson) {
   };
 }
 
+/**
+ * @param {Options} options
+ * @param {string} dest
+ * @param {string} elmModulePath
```
jfmengels, line 363: I think this would be more accurate:
```
 * @param {Path} dest
 * @param {Path} elmModulePath
```

lib/remote-template.js
lishaduck: They should, but I used the IDE inferer to get started, which skips them, unfortunately, and I didn't realize it until I was done.
TS infers them though, so I think it can be left for a follow-up.

jfmengels:
Woohoo! This is a great step forward!
Thank you @lishaduck :blush:

PR #302, opened 2024-11-06
Better TS

lishaduck:
Follow up to #301.

Works on TSESlint-based safety.
Probably broke some stuff with error handling.

Please review closely,[^1] some of these changes might be better served with ignores. Then again, ignores are evil, so... I just fixed them all and hoped for the best.

Oh, and I added more JSDocs for literally everything! Yay!

[^1]: EDIT: 🤣

### DIFF
--- a/lib/elm-app-worker.js
+++ b/lib/elm-app-worker.js
@@ -1,12 +1,20 @@
 /**
- * @import {MessagePort} from 'worker_threads';
+ * @import {MessagePort} from 'node:worker_threads';
  * @import {Ports} from './types/app';
+ * @import {WorkerData} from './types/elm-app-wrapper';
  * @import {SendPort} from './types/promisify-port';
+ * @import {WorkerThreads} from './types/worker';
  */
-const {parentPort, workerData} = require('node:worker_threads');
+
+const workerThreads = require('node:worker_threads');
 const ElmCommunication = require('./elm-communication');
 const loadCompiledElmApp = require('./load-compiled-app');
 const ResultCache = require('./result-cache');
+
+/**
+ * @type {WorkerThreads<WorkerData>}
+ */
+const {parentPort, workerData} = workerThreads;
 
 const elmModule = loadCompiledElmApp(workerData.elmModulePath);
 
--- a/lib/options.js
+++ b/lib/options.js
@@ -101,6 +101,9 @@ function compute(processArgv) {
     }
   }
 
+  /**
+   * @returns {Path}
+   */
   function projectToReview() {
     if (!elmJsonPath) {
       throw new ErrorMessage.CustomError(
@@ -142,6 +145,9 @@ try re-running it with ${chalk.cyan('--elmjson <path-to-elm.json>')}.`,
 
   const namespace = args.namespace || 'cli';
 
+  /**
+   * @returns {Path}
+   */
   function elmStuffFolder() {
     return path.join(
       projectToReview(),
@@ -154,12 +160,16 @@ try re-running it with ${chalk.cyan('--elmjson <path-to-elm.json>')}.`,
     );
   }
 
+  /**
+   * @returns {Path}
+   */
   function userSrc() {
     return args.config
       ? path.resolve(process.cwd(), args.config)
       : path.join(projectToReview(), 'review');
   }
 
+  /** @returns {Path} */
   function suppressedErrorsFolder() {
     return path.join(userSrc(), 'suppressed');
   }
@@ -518,7 +528,11 @@ function suggestions(flagName) {
       distance: levenshtein.distance(flagName, f.name)
     }))
     .sort(
-      /** @returns {-1 | 0 | 1} */
+      /**
+       * @param {Flag & {distance: number}} a
+       * @param {Flag & {distance: number}} b
+       * @returns {-1 | 0 | 1}
+       */
       (a, b) => {
         if (a.distance < b.distance) {
           return -1;
@@ -542,7 +556,10 @@ function suggestions(flagName) {
  */
 function checkForMissingArgs(subcommand, args) {
   for (const flag of Flags.flags.filter(
-    /** @returns {flag is Flag & Multi} */
+    /**
+     * @param {Flag} flag
+     * @returns {flag is Flag & Multi}
+     */
     (flag) => flag.boolean === false
   )) {
     if (typeof args[flag.name] === 'boolean') {
--- a/lib/options.js
+++ b/lib/options.js
@@ -181,8 +181,10 @@ try re-running it with ${chalk.cyan('--elmjson <path-to-elm.json>')}.`,
 
   const localElmReviewSrc = process.env.LOCAL_ELM_REVIEW_SRC;
 
-  const [authorName, packageName, license] = args.prefill
-    ? args.prefill.split(',')
+  /** @type {string} */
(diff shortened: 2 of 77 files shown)


lib/utils.js
```diff
@@ -7,6 +7,19 @@ function unique(array) {
...
+/**
+ * Convert anything into an error.
+ *
+ * If it's already an error, do nothing.
+ * Otherwise, wrap it in a generic error.
+ *
+ * @param {unknown} error
+ */
+function intoError(error) {
+  return error instanceof Error ? error : new Error(String(error));
```
jfmengels, line 20: I'm curious, did you notice places where we (potentially) wrapped an exception into multiple `Error`s? (nested I mean)

lib/elm-app-worker.js
```diff
@@ -1,13 +1,21 @@
...
-const {parentPort, workerData} = require('node:worker_threads');
+
+const workerThreads = require('node:worker_threads');
 const ElmCommunication = require('./elm-communication');
 const loadCompiledElmApp = require('./load-compiled-app');
 const ResultCache = require('./result-cache');
 
+/**
+ * @type {WorkerThreads<WorkerData>}
+ */
```
jfmengels, line 16: Should the type annotation not ideally be placed at the definition of `workerThreads`? (That at least sounds a lot more intuitive to me)

lib/options.js
```diff
@@ -50,22 +50,23 @@ function compute(processArgv) {
     processArgv.slice(2).includes('--help') ||
     processArgv.slice(2).includes('-h');
 
-  /** @type {ParsedArgs} */
-  const args = minimist(processArgv.slice(2), {
-    alias: Flags.flags.reduce((object, flag) => {
-      if (flag.alias !== undefined) {
-        object[flag.name] = flag.alias;
-      }
+  const args = /** @type {ParsedArgs} */ (
```
jfmengels, line 53: What's the difference between
```
const args = /** @type {ParsedArgs} */ ...
```
and
```
/** @type {ParsedArgs} */
const args = ...
```
Are the two different from TS' point of view?
(Or is this just a styling change?)

lib/remote-template.js
```diff
@@ -338,7 +340,10 @@ async function makeGitHubApiRequest(options, url, handleNotFound) {
     // TODO(@lishaduck) [engine:node@>=21]: We can use `fetch` now.
     const response = await got(url, parameters);
 
-    return response.body;
+    /** @type {{body: unknown}} */
```
jfmengels, line 345: Seeing this over and over, I feel like `TS` should try to understand type annotations on a return statement like
```ts
/** @type {{body: unknown}} */
return response.body;
```

lib/template-dependencies.js
```diff
@@ -119,21 +119,29 @@ function solve(options, elmVersion, elmJson, extra, onlineFirst) {
...
-        : dependencyProvider.solveOffline(options, elmVersion, elmJson, extra)
+    const dependencies = /** @type {unknown} */ (
+      JSON.parse(
+        onlineFirst && !options.offline
+          ? dependencyProvider.solveOnline(options, elmVersion, elmJson, extra)
+          : dependencyProvider.solveOffline(options, elmVersion, elmJson, extra)
+      )
     );
+
+    return /** @type {ApplicationDependencies} */ (dependencies);
```
jfmengels, line 130: Just so I understand, the `/** @type {unknown} */` is for correctly annotating the value of `dependencies`, and the type annotation here is to cast?

Or are they both for casting? In which case, why not cast only once?

lib/project-dependencies.js
```diff
@@ -29,7 +30,8 @@ async function collect(options, elmJson, elmVersion) {
   let hasDependenciesThatCouldNotBeDownloaded = false;
 
   const projectDepsPromises = Object.entries(dependenciesEntries).map(
-    async ([name, constraint]) => {
+    async ([pkgname, constraint]) => {
+      const name = /** @type {PackageName} */ (pkgname);
```
jfmengels, line 34: (Mostly out of curiosity) Could this type potentially be inferred by typing `dependenciesEntries` instead?

test/jest-helpers/cli.js
```diff
@@ -12,15 +12,16 @@ expect.extend({toMatchFile});
 
 /**
  * @param {string} args
- * @param {Options} [options=undefined]
+ * @param {Options} [options]
+ * @returns
```
jfmengels, line 16: Is the type not missing here?

lib/runner.js
```diff
@@ -139,7 +139,7 @@ async function initializeApp(options, elmModulePath, reviewElmJson, appHash) {
      * @returns {never}
      */
     (errors) => {
-      Report.report(options, {errors});
+      void Report.report(options, {errors});
```
jfmengels, line 142: I'm not sure I knew that `worker.terminate()` returned a `Promise` back when I wrote it, but I think all the floating promises in this commit make sense to keep as floating (for concurrency purposes, etc.).

The only one I'd be worried about it potentially this one, but so far it has always worked well so... I think it's fine to keep as is :shrug:

jfmengels:
Great work!

Btw, thank you for your nice commit etiquette for your PRs. It's easy to review each commit (except maybe for 3965c34b3c9339a6c77cc6654e69e375e22684ad :stuck_out_tongue: , although it was not that bad in practice)

lib/utils.js
lishaduck: > I'm curious, did you notice places where we (potentially) wrapped an exception into multiple `Error`s? (nested I mean)

Nope.

> Related: did you notice places where we throw non-Errors? 🤔 (Note, I haven't read the rest of the commits yet)

Yeah, we throw strings in a few places.

lib/options.js
lishaduck: The first one is a cast, the second one is a type annotation. I don't recall why this needed a cast.

lib/remote-template.js
lishaduck: It's actually a ts-eslint issue: typescript-eslint/typescript-eslint#1682. It can't read JSDoc casts b/c they're technically just comments, so the type queries are wrong. `no-unsafe-return` bans returning `any`, which is good, but can't read immediate casts, it needs an intermediate variable. In normal TS, you can just use `as`.

lib/project-dependencies.js
lishaduck: It is typed, but because of excess properties, ts can't infer it: [FAQ § Excess Properties Are OK](https://github.com/microsoft/TypeScript/wiki/FAQ#indirect-excess-properties-are-ok).

lib/template-dependencies.js
lishaduck: When using `JSON.parse` (& other lib functions), you generally want to cast to unknown, then to another type (the `as unknown as` pattern), because they don't default to `unknown` for back compat.
I'd guess it's not a direct return b/c of `no-unsafe-returns`.

test/jest-helpers/cli.js
lishaduck: With options, the `[]` syntax defaults to undefined.
With the `@returns`, I must've just missed it.

lib/elm-app-worker.js
lishaduck: I tried that, but it doesn't like it.

PR #303, opened 2024-11-10
Remove fs-extra dependency

lishaduck:
Manual rebase of #147.

Fixes: #120.
Supersedes/closes: #147.

### DIFF
--- a/lib/init.js
+++ b/lib/init.js
@@ -5,8 +5,6 @@
 
 const path = require('node:path');
 const fs = require('graceful-fs');
-// TODO(@lishaduck) [engine:node@>=16.7.0]: Use native `cp`.
-const fsExtra = require('fs-extra');
 const chalk = require('chalk');
 const prompts = require('prompts');
 const FS = require('./fs-wrapper');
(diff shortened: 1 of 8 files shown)


lib/init.js
```diff
@@ -6,7 +6,6 @@
 const path = require('node:path');
 const fs = require('graceful-fs');
 // TODO(@lishaduck) [engine:node@>=16.7.0]: Use native `cp`.
-const fsExtra = require('fs-extra');
```
jfmengels, line 9: The comment above is now outdated, no?

lib/new-package.js
```diff
@@ -455,11 +461,14 @@ function packageJson(options, packageName) {
       'update-examples': 'node maintenance/update-examples-from-preview.js',
       postinstall: 'elm-tooling install'
     },
-    dependencies: {
+    engines: {
+      node: '>=14.21.3'
+    },
+    devDependencies: {
```
jfmengels, line 467: I'd keep these as dependencies. Elm packages that use `package.json` tend to have these as regular dependencies. Though I'm not sure that one is better than the other :shrug:

lishaduck: I'd argue they're better as dependencies, but I'd meant to move these to another PR anyway.

jfmengels:
One less dependency! Thank you :rocket:

PR #307, opened 2024-11-10
Even Better TS

lishaduck:
Follow up to #302 (and #186 and #187).


lib/sync-get-worker.js
```diff
@@ -5,7 +5,7 @@
 
 const https = require('node:https');
 const workerThreads = require('node:worker_threads');
-const {result} = require('./core');
+const {Result} = require('./core');
```
jfmengels, line 8: I'd be happy with a `require('./result')` to avoid the evil barrel export (and even so in general). What do you think?

.github/workflows/test.yml
```diff
@@ -38,6 +38,10 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - name: Git checkout
+      # Pin actions their full commit hashes to prevent supply-chain attacks.
+      # To update, find the latest tag on the releases page,
+      # then copy the full SHA and paste it back here (and update the comment).
+      # Alternatively, setup Dependabot for the `github-actions` ecosystem.
```
jfmengels, line 44: Thank you for circling back to this :heart:

lib/sync-get-worker.js
lishaduck: I agree! 😀

lishaduck:
Hit rate limits. Will try again later.

jfmengels (approved):

PR #308, opened 2024-11-10
Update generated `package.json`s

lishaduck:
Trying to make the generated `package.json`s more accurate.

Extracted from #303.

### DIFF
--- a/lib/new-package.js
+++ b/lib/new-package.js
@@ -308,6 +308,7 @@ ElmjutsuDumMyM0DuL3.elm
   const licenseArgs = options.forTests ? '--name "Test User" --year 2020' : '';
   try {
     childProcess.execSync(
+      // TODO(@lishaduck): Evaluate calling the API instead.
       `npx license ${license} --projectName "${authorName}/${packageName}" ${licenseArgs}`,
       {
         cwd: dir,
@@ -445,6 +446,7 @@ elm-review --template ${authorName}/${packageName}/example
 function packageJson(options, packageName) {
   return {
     name: packageName,
+    private: true,
     scripts: {
       test: 'npm-run-all --print-name --silent --sequential test:make test:format test:run test:review test:package',
       'test:make': 'elm make --docs=docs.json',
@@ -459,11 +461,14 @@ function packageJson(options, packageName) {
       'update-examples': 'node maintenance/update-examples-from-preview.js',
       postinstall: 'elm-tooling install'
     },
-    dependencies: {
+    engines: {
+      node: '>=14.21.3'
+    },
+    devDependencies: {
       'elm-doc-preview': '^5.0.5',
       'elm-review': `^${options.packageJsonVersion}`,
-      'elm-test': '^0.19.1-revision10',
-      'elm-tooling': '^1.13.1',
+      'elm-test': '^0.19.1-revision12',
+      'elm-tooling': '^1.15.1',
       'fs-extra': '^9.0.0',
       glob: '^9.3.1',
       'npm-run-all': '^4.1.5'
--- a/test/run-snapshots/elm-review-something-for-new-rule/package.json
+++ b/test/run-snapshots/elm-review-something-for-new-rule/package.json
@@ -1,5 +1,6 @@
 {
   "name": "elm-review-something",
+  "private": true,
   "scripts": {
     "test": "npm-run-all --print-name --silent --sequential test:make test:format test:run test:review test:package",
     "test:make": "elm make --docs=docs.json",
@@ -13,11 +14,14 @@
     "update-examples": "node maintenance/update-examples-from-preview.js",
     "postinstall": "elm-tooling install"
   },
-  "dependencies": {
+  "engines": {
+    "node": ">=14.21.3"
+  },
+  "devDependencies": {
     "elm-doc-preview": "^5.0.5",
     "elm-review": "^2.12.0",
-    "elm-test": "^0.19.1-revision10",
-    "elm-tooling": "^1.13.1",
+    "elm-test": "^0.19.1-revision12",
+    "elm-tooling": "^1.15.1",
     "fs-extra": "^9.0.0",
     "glob": "^9.3.1",
     "npm-run-all": "^4.1.5"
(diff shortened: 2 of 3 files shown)


lishaduck:
Awww. Hit rate limits. :(

lib/new-package.js
```diff
@@ -305,6 +305,7 @@ ElmjutsuDumMyM0DuL3.elm
   const licenseArgs = options.forTests ? '--name "Test User" --year 2020' : '';
   try {
     childProcess.execSync(
+      // TODO(@lishaduck): Evaluate calling the API instead.
```
lishaduck, line 308: Pros:
- Direct call (perf)
- Clearer dependency chain
- Works even if npx isn't on path

Cons:
- Would show up as unmaintained on socket, etc
- Depends on deprecated packages
- Bloats `node_modules` for 99% of users

lib/new-package.js
```diff
@@ -455,11 +457,14 @@ function packageJson(options, packageName) {
       'update-examples': 'node maintenance/update-examples-from-preview.js',
       postinstall: 'elm-tooling install'
     },
-    dependencies: {
+    engines: {
+      node: '>=14.21.3'
+    },
```
lishaduck, line 462: I assume we want the same range as `elm-review` itself, so we can test it.

lib/new-package.js
```diff
@@ -455,11 +457,14 @@ function packageJson(options, packageName) {
...
+    devDependencies: {
```
lishaduck, line 463: If they're only needed at development-time, I don't see why we should need them as dependencies, but if you'd rather revert them, that's fine as well.

lib/new-package.js
```diff
@@ -455,11 +457,14 @@ function packageJson(options, packageName) {
...
       'elm-doc-preview': '^5.0.5',
       'elm-review': `^${options.packageJsonVersion}`,
       'elm-test': '^0.19.1-revision10',
-      'elm-tooling': '^1.13.1',
+      'elm-tooling': '^1.15.1',
```
lishaduck, line 467: Bumped `elm-tooling` to latest.

lib/new-package.js
```diff
@@ -455,11 +457,14 @@ function packageJson(options, packageName) {
...
       postinstall: 'elm-tooling install'
     },
-    dependencies: {
+    engines: {
+      node: '>=14.21.3'
+    },
+    devDependencies: {
       'elm-doc-preview': '^5.0.5',
       'elm-review': `^${options.packageJsonVersion}`,
       'elm-test': '^0.19.1-revision10',
```
lishaduck, line 466: Missed this one.

lib/new-package.js
```diff
@@ -455,11 +457,14 @@ function packageJson(options, packageName) {
...
-      'elm-tooling': '^1.13.1',
+      'elm-tooling': '^1.15.1',
       'fs-extra': '^9.0.0',
```
lishaduck, line 468: Leaving this node 14 compatible.

lib/new-package.js
```diff
@@ -455,11 +457,14 @@ function packageJson(options, packageName) {
...
       glob: '^9.3.1',
```
lishaduck, line 469: Will probably switch to `tinyglobby` in a followup.

lib/new-package.js
```diff
@@ -455,11 +457,14 @@ function packageJson(options, packageName) {
...
       'npm-run-all': '^4.1.5'
```
lishaduck, line 470: Turbo's probably overkill, `npm-run-all2` requires newer node, so I'm leaving as-is. If we find people switching because bugs surface, we can change it, but I think it's fine for now.

lishaduck:
I've noticed this there's a test that's been flaking ("Running using other configuration (without errors)"):

```console

  https://github.com/jfmengels/node-elm-review/issues/new

Please include this error message and as much detail as you can provide. Running
with --debug might give additional information. If you can, please provide a
setup that makes it easy to reproduce the error. That will make it much easier
to fix the issue.

<local-path>/test/project-with-errors/elm-stuff/generated-code/jfmengels/elm-review/cli/<version>/review-applications/bd2a2bb6c965dac5241ce8c4d0615d83.js:10328
    		$temp$offset = offset + 
    						                    

SyntaxError: Unexpected end of input

  at loadCompiledElmApp (<local-path>/lib/load-compiled-app.js:28:18)
  at initWithoutWorker (<local-path>/lib/app-wrapper.js:144:21)
  "
  at Object.test (test/review.test.js:37:1)

```

lib/new-package.js
lishaduck: Got it :)

lib/new-package.js
jfmengels: We could also inline the functionality of the package: support only the common ones and let the user opt-out if they want a different one.

jfmengels:
> I've noticed this there's a test that's been flaking ("Running using other configuration (without errors)"):

Yes, I've noticed the same thing. I had a look, thinking it could be us not waiting for the compilation to be finalized, but the code looks fine. I still think it's the writing of the file (or the loading of it, but that's weirder) that is getting cut off early somehow, but I haven't figured out how yet. :thinking: 

It sounds pretty recent though. I don't think we saw this one before, but I could be wrong.

lib/new-package.js
```diff
@@ -308,6 +308,7 @@ ElmjutsuDumMyM0DuL3.elm
   const licenseArgs = options.forTests ? '--name "Test User" --year 2020' : '';
```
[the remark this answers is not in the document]
lishaduck: 2020 is for tests, otherwise we leave it up for the user.

lishaduck:
> It sounds pretty recent though. I don't think we saw this one before, but I could be wrong.

Well, it's really started flaking. Will look into it.
For reference, here's the first failure from it: https://github.com/jfmengels/node-elm-review/actions/runs/11769034390/job/32779541170

lishaduck:
Was able to repro with 38453c161ca3c4273482760b7c4576203e3ced72, so sadly not #303.

lishaduck:
@jfmengels, any ideas?

PR #309, opened 2024-11-10
Replace GITHUB_AUTH by AUTH_GITHUB

jfmengels:
Environment variables in GitHub can't start with "GITHUB_".

I created an environment for GitHub Actions with a `AUTH_GITHUB`, which will be used for tests that use `--template`. Hopefully this will help prevent rate limits :crossed_fingers:

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -14,14 +14,14 @@ replace_script() {
 }
 
 # If you get errors like rate limit exceeded, you can run these tests
-# with "GITHUB_AUTH=gitHubUserName:token"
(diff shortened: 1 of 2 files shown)


jfmengels:
I don't know if it worked, I guess we'll see as we do more changes :man_shrugging:

lishaduck:
It worth considering that now it'll directly count against you (i.e., if say I wrote some code to spam the api, you'd be the one flagged for abuse), perhaps you should use `secrets.GITHUB_TOKEN` (built-in token, granular permissions), which would count against the triggerer.

PR #310, opened 2024-11-11
Use GHA's token

### LINKED ISSUE #336: Resupport GH Auth
I looked into this some in #310, and I'm pretty certain our workflows aren't set up correctly.

First off, we never pass the secret into the environment. #310 fixes that.
Next off, we're using username:token. Why do we need a username? If we pass in a token, the API knows who's associated with that. Moreover, the code uses "password" for token, leading me to think that perhaps it was from when the API supported password auth?
Then, we should also specify the version of the REST API we want, which will prevent breaking changes.
Oh, and we're passing the token in using HTTP Basic Auth which a) is very insecure and b) is only support for GitHub Apps' tokens, not user tokens.

So... I think that's why we hit rate limits a lot.
I think #310 covers it all, ~~but it's failing CI on a genuine error, last I checked (I should try rebasing once main is green again)~~ EDIT: rebased, green, seems to work well.

lishaduck:
GHA provides a token itself, we don't need to generate one. This is more-fine grained, easier to change if needed, and more secure.

I think the existing `contents: read` is sufficient, but it might not be. We'll see.

Based on https://github.com/jfmengels/node-elm-review/pull/309#issuecomment-2467063929.

Closes #336.

### DIFF
--- a/lib/remote-template.js
+++ b/lib/remote-template.js
@@ -197,21 +197,31 @@ ${error.message}`
   }
 }
 
-/** @type {ErrorMessageInfo} */
-const rateLimitErrorMessage = {
-  title: 'GITHUB RATE LIMIT EXCEEDED',
-  message: `It looks like you exceeded the GitHub rate limit by using "--template" too many
-times, this will likely last for 30 minutes.
-
+/**
+ * @param {boolean} hasPat
+ * @param {string} resetTime
+ * @returns {ErrorMessageInfo}
+ */
+function rateLimitErrorMessage(hasPat, resetTime) {
+  return {
+    title: 'GITHUB RATE LIMIT EXCEEDED',
+    message: `It looks like you exceeded the GitHub rate limit by using "--template" too many
+times, this will likely last until ${resetTime}.
+${
+  hasPat
+    ? ''
+    : `
 In the meantime, you can use \`--github-auth=your-api-token\`.
-Follow this guide: https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token to create an API token, and give it access to public repositories.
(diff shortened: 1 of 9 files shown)


lishaduck:
Oh, github-auth needs a username. I'll think on this.

lishaduck:
Ok, so the GITHUB_TOKEN is for a builtin github app, not your own account, so that i.e., contributing to a project with extensive doesn't use up your personal rate. Thus, there's not a username associated with it. However, I think we don't need a username to make authenticated requests, do we? Could we make it optional?

lishaduck:
Hmm, looks like it's not picking up AUTH_GITHUB.

lishaduck:
Hit #313.

lishaduck:
I'll try to rebase this when I get a moment, I still think this should fix rate limiting (because secrets isn't injected into ENV), and still obsolete needing @jfmengels' token.

lib/remote-template.js
```diff
@@ -197,21 +197,31 @@ ${error.message}`
...
+/**
+ * @param {boolean} hasPat
+ * @param {string} resetTime
+ * @returns {ErrorMessageInfo}
+ */
+function rateLimitErrorMessage(hasPat, resetTime) {
+  return {
+    title: 'GITHUB RATE LIMIT EXCEEDED',
+    message: `It looks like you exceeded the GitHub rate limit by using "--template" too many
+times, this will likely last until ${resetTime}.
```
jfmengels, line 209: This is so nice to have!

jfmengels:
This looks to work promisingly.

lishaduck:
It passes CI now!

PR #281, opened 2024-10-30
Swap to tinyglobby

lishaduck:
I was testing if it worked on 14 only to realize I'd written the PR and might as well send it in.

Blocked on: #290.

Issue #223, opened 2024-07-19
Leaner Dependencies


lishaduck:
We use (`node-`)`glob`, `globby`, and (as a dependency of `globby`) `fast-glob`. We could probably cut down to just `fast-glob`, probably. Almost certainly droping `glob` would be good.
Apparently people have been switching off of Chalk recently in favor of Picocolors.
`node-fetch-native` is much leaner than `node-fetch` (though we use `got`, don't we?), and defaults to native fetch in Node 21+.
We're also using `cross-spawn`, many projects switched to Execa a while back, but because it's ESM-only, but I recently saw a few projects switch to `@jsdevtools/ez-spawn` as a more correct, smaller, and faster CommonJS `cross-spawn`/Execa.

Globbing's been been on the todo list for a bit, figured I'd file an issue.

lydell:
Regarding cross-spawn, there’s also this: https://github.com/lydell/elm-watch/pull/66. The key here is that we only need to support spawning the different elm tools, not stuff in general, which makes the problem space much smaller. Basically, the approach is to first try to spawn `my-command` and if that fails (on Windows only) try spawning `my-command.cmd`. I don’t remember why I haven’t merged that though…

lydell:
cross-spawn, glob and chalk are also used in node-test-runner. I _think_ there is a big overlap in users between node-test-runner and elm-review, so for the dependencies to _actually_ get leaner for many people, we need to make upgrades there as well (which I have nothing against, I love fewer dependencies, the only road block is fear of breakage).

lishaduck:
> Regarding cross-spawn, there’s also this: [lydell/elm-watch#66](https://github.com/lydell/elm-watch/pull/66). The key here is that we only need to support spawning the different elm tools, not stuff in general, which makes the problem space much smaller. Basically, the approach is to first try to spawn `my-command` and if that fails (on Windows only) try spawning `my-command.cmd`. I don’t remember why I haven’t merged that though…

[...]

PR #311, opened 2024-11-11
Replace glob with tinyglobby

lishaduck:
Part of #223.

Follow-up to #281.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -177,6 +177,7 @@ module.exports = {
       rules: {
         'n/no-process-exit': 'off',
         'n/no-missing-require': 'off', // `require` of `elm.json`.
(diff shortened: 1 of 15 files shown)


.eslintrc.js
```diff
@@ -177,6 +177,7 @@ module.exports = {
       rules: {
         'n/no-process-exit': 'off',
         'n/no-missing-require': 'off', // `require` of `elm.json`.
+        'n/no-extraneous-require': ['error', {allowModules: ['fs-extra']}],
```
jfmengels, line 180: This is for files generated by `new-package`, correct?

lishaduck: Yeah. Not sure this issue only surfaces with these changes, but it should've been in #303.

PR #321, opened 2024-11-29
Include files from a source-directory if that source-directory is in elm-stuff

### LINKED ISSUE #320: NoUnused.Variables reports wrong alias as unused when importing module from elm-git-install dependency
**Describe the bug**
When importing two modules as the same alias, and one of the modules comes from a package installed by elm-git-install, NoUnused.Variables reports the used module as unused, and the unused as used.

**SSCCE (Short Self-Contained Correct Example)**
1. Clone https://github.com/morteako/elm-review-test-case : `git clone git@github.com:morteako/elm-review-test-case.git`
1.5. optional: The results from running elm-git-install is already included in the repo, but
if one wants or needs (if issues) to actually run `elm-git-install`, just run `rm -rf elm-stuff && npx elm-git-install` and it will overwrite the contents if needed.
2. Run `npx elm-review`
3. See wrong first error and correct second error
4. Check out source code in src/ and https://github.com/morteako/elm-review-test-case/blob/master/elm-stuff/gitdeps/github.com/morteako/elm-review-test-case-git-install.git/src/ if one would like

**Expected behavior**
I expect elm-review and NoUnused.Variables to report the correct module alias as unused, namely: `import UtilNotUsed as List`

**Additional context**
Code that triggers the bug, where Util.List is installed in elm-stuff, using elm-git-install
```elm
module MainGitInstallWrongError exposing (..)

-- actually used, reported as unused
import Util.List as List

-- actually unused, not reported as unused
import UtilNotUsed as List

useImportAliasFromElmstuff x = List.test x
```

The actually unused module does not contain a `test`-export
```elm
module UtilNotUsed exposing (x)

x = 1
```

The module installed using elm-git-install can be seen here:
It does export `test` and is the therefore the module alias that is actually used
https://github.com/morteako/elm-review-test-case/blob/master/elm-stuff/gitdeps/github.com/morteako/elm-review-test-case-git-install.git/src/Util/List.elm

and contains 
```elm
module Util.List exposing (test)

test : a -> a
test x = x
```

To see full code, see [repo](https://github.com/morteako/elm-review-test-case)

I can't reproduce the bug when import two local modules instead (see https://github.com/morteako/elm-review-test-case/blob/master/src/MainLocalCorrectError.elm), so it seems to be related to elm-git-install and the fact that it places the code in elm-stuff/gitdeps/github.com/...

I have not had any other bugs like this related to elm-git-install packages and elm-review though.

jfmengels:
Fixes #320 reported by @morteako

We now only ignore `elm-stuff` inside every `source-directory`, rather than all of them.
This includes a test that makes sure that those files get included in the analysis AND that we can report errors for them.


lishaduck:
Hm. The failing test isn't the flake. I assume it passes locally?

jfmengels:
Correct, yes. Well, at least that new test passes. This is symptomatic of `elm-stuff` being ignored/absent in CI (because that's the error I was trying to fix) :thinking:

jfmengels:
Oh, it's because a file did not committed. I somehow made an exclusion to it in `.gitignore`, but it still doesn't want to get included. I'm investigating why.

lishaduck:
> Oh, it's because a file did not committed. I somehow made an exclusion to it in `.gitignore`, but it still doesn't want to get included. I'm investigating why.
> 
> EDIT: Success. I couldn't get it to be included by editing the `.gitignore` file (I still don't understand why), but I forced the file to be added by doing `git add <file> -f`.

If I recall correctly, you can't unignore folder contents if the folder is ignored, so rather than blanket ignoring `elm-stuff`, you'd want to ignore `**/elm-stuff/**` (and you'd need the leading `**/` because it's a non-trivial pattern—we love glob syntax 🤣)

morteako:
@jfmengels Seems like this also fixes the same issue but for NoUnused.Dependencies (which makes sense), which is great 💯

jfmengels:
> If I recall correctly, you can't unignore folder contents if the folder is ignored, so rather than blanket ignoring elm-stuff, you'd want to ignore `**/elm-stuff/**` (and you'd need the leading `**/` because it's a non-trivial pattern—we love glob syntax 🤣)

Oh right... I knew that because I modeled the extra files API based on that, but my version makes that a bit more explicit and I forgot about the original :sweat_smile: 
But something weird happened locally, because even by removing the `elm-stuff` line, it didn't ignore `elm-stuff`. I don't know what happened but problem solved so :shrug: 

> Seems like this also fixes the same issue but for NoUnused.Dependencies (which makes sense), which is great 💯

That's great @morteako! I'll try to publish in a few days, I hope this is not blocking you (please tell me if it is).

Issue #319, opened 2024-11-29
Error on first time running elm-pages


arkaitz-dev:
elm-pages dev
elm-pages dev server running at <http://localhost:1234>
Ran in 2ms   /elm.js
Ran elm-review in 252ms
-- UNEXPECTED ERROR --------------- /home/arkaitz/.nvm/versions/node/v20.18.0/lib/node_modules/elm-pages/generator/review/elm.json
I ran into an unexpected error. Please open an issue at the following link:
  https://github.com/jfmengels/node-elm-review/issues/new

Please include this error message and as much detail as you can provide. If you
can, please provide a setup that makes it easy to reproduce the error. That will
make it much easier to fix the issue.

Below is the error that was encountered.
--------------------------------------------------------------------------------
/home/arkaitz/proyectos/elm/prueba_pages/elm-stuff/generated-code/jfmengels/elm-review/elm-pages/2.12.0/review-applications/59274e92599855caa16357e0a28a68ee.js:6089
var $jfmengels$elm_review$Review$Exceptions$isFileWeWantReportsFor = F2(function $jfm);
                                                                                     ^

SyntaxError: Unexpected token ')'
    at wrapSafe (node:internal/modules/cjs/loader:1378:20)
    at Module._compile (node:internal/modules/cjs/loader:1428:41)
    at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module._load (node:internal/modules/cjs/loader:1104:12)
    at Module.require (node:internal/modules/cjs/loader:1311:19)
    at require (node:internal/modules/helpers:179:18)
    at loadCompiledElmApp (/home/arkaitz/.nvm/versions/node/v20.18.0/lib/node_modules/elm-review/lib/load-compiled-app.js:28:10)
    at initWithoutWorker (/home/arkaitz/.nvm/versions/node/v20.18.0/lib/node_modules/elm-review/lib/app-wrapper.js:110:21)
    at init (/home/arkaitz/.nvm/versions/node/v20.18.0/lib/node_modules/elm-review/lib/app-wrapper.js:35:10)
Trace: problem with format in reviewReportJsonString TypeError: Cannot read properties of undefined (reading 'length')
    at emptyReviewError (file:///home/arkaitz/.nvm/versions/node/v20.18.0/lib/node_modules/elm-pages/generator/src/dev-server.js:613:55)
[...]

PR #322, opened 2024-12-02
Set `esModuleInterop` to true

lishaduck:
I turned it off a while ago, but it should be on. It's `allowSyntheticDefaultImports` that should be off.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "prettier-check": "prettier . --check --cache",
     "prettier-fix": "prettier . --write --cache",
     "test": "turbo run testing check-engines --continue",
+    "test-sync": "npm run test-run && npm run jest",
     "test-run": "(cd test/ && ./run.sh)",
     "test-run-record": "(cd test/ && ./run.sh record)",
     "tsc": "tsc",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,7 @@
     "moduleResolution": "Node16",
     "target": "es2020",
     "lib": ["ES2020"],
-    "esModuleInterop": false,
+    "esModuleInterop": true,
     "allowSyntheticDefaultImports": false,
     "noEmit": true,
     "noImplicitThis": true,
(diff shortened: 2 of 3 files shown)


lishaduck (commented):
Ok, ran the following the my terminal twice, no flakes at all: 

```bash
for ((i=1; i<=5; i++)); do
turbo --ui=stream --force=true jest
done
```

Makes me wonder if the flake is Linux exclusive. I know I reproed it locally a while ago, but I can't seem to anymore.

jfmengels:
Hmm possible. I managed to recreate similar flakes locally (Linux Ubuntu).

jfmengels:
I was only able to recreate it by having multiple terminals running `jest`. So maybe it's because two tests try to compile the same application at the same time? :thinking:

lishaduck:
> I was only able to recreate it by having multiple terminals running `jest`. So maybe it's because two tests try to compile the same application at the same time? 🤔

Hmm, that'd explain #319, as `elm-pages` has a lot of race conditions (I hear it's a lot better nowadays, so maybe not).
However, Jest tests don't run in parallel, do they? Oh... e2e tests and jest tests run in parallel... I'll try reverting that :)

lishaduck:
> I was only able to recreate it by having multiple terminals running `jest`. So maybe it's because two tests try to compile the same application at the same time? 🤔

I, too, can repro it locally like that.

lishaduck:
Disappointing, but I guess I just means I need to find the time to do that refactor of options parsing that'll let us quit spawning processes.

lishaduck:
@jfmengels, do you want me to file an issue for supporting concurrent `elm-review` runs?

jfmengels:
Yeah, the namespace option was meant for that. Though it's also a shame to need that, as that means that a cache for the CLI would not get used by an editor and vice versa. If we could get rid of it that would actually be nice I think.

@lishaduck opening an issue would be great 👍

Issue #176, opened 2024-06-25
Testing Improvements


lishaduck:
- [ ] Don't make tests need internet/GHA API
  - [ ] Migrate to `octokit`?[^1] (we get types to replace tests) (Node 18, #171-only)
- [ ] Fix flag parsing and enable memory leak tracking (https://github.com/jfmengels/node-elm-review/issues/156#issuecomment-2184082300)
- [x] Fix "error: conflicting output format options" from `diff` (#193)
- [x] #197
- [x] #224
- [ ] Don't break snapshots on new Elm package releases (ref: #185) 🌟
  - [ ] We could perhaps set up a cron job to automate it until we get a better solution?
- [ ] Maybe migrate to Vitest, as it's way faster, has a better ide integration, and supports #171?
- [ ] #378
- [x] #313
- [ ] Make tests agnostic of the cwd.
- [ ] #343
- [ ] 100% Code Coverage (stretch goal; requires not shelling out)

[^1]: Alternatively, `ungh` would help with rate limiting and types once `ofetch` finishes adding in support for statically typing API endpoints (https://github.com/unjs/ungh/issues/4#issuecomment-1401762324).

lishaduck:
I don't know if this is visible without a GitKraken account, but here's a patch: https://gitkraken.dev/link/drafts/ba02a3a7-4beb-4ed9-aac8-447540488f70?type=patch

lishaduck:
Yay! One of my issues was "yet another longstanding NPM bug™," so it can be removed from this list, at least.

jfmengels:
> I don't know if this is visible without a GitKraken account, but here's a patch for offline testing for now:

It is not, unfortunately.

lishaduck:
> > I don't know if this is visible without a GitKraken account, but here's a patch for offline testing for now:
> 
> It is not, unfortunately.

I figured. Why do they have to overwrite copy-as-patch? Well, there's the actual patch:
```diff
diff --git a/test/run.sh b/test/run.sh
index 86bccb7..1c38065 100755
--- a/test/run.sh
+++ b/test/run.sh
@@ -163,9 +163,9 @@ else
   echo -e '\x1B[33m-- Testing runs\x1B[0m'
 fi
 
-PACKAGE_PATH=$(npm pack -s ../ | tail -n 1)
+PACKAGE_PATH=$(npm pack --offline -s ../ | tail -n 1)
 echo "Package path is $PACKAGE_PATH"
[...]

PR #325, opened 2024-12-04
Replace shell scripts

lishaduck:
Begin working toward the goals in #176.

VERY WIP!!!

After all that work trying to fix the shell, I decided to just nuke it and get copilot to do a basic js rewrite, which exposed a bunch of bugs. This simplifies the code at a (likely) small speed decrease.

---

Eventually (not in this PR 🤞), I'd like to replace these with test cases in Jest; even more eventually, I'd like to quit spawning processes in Jest tests, which would allow us to mock the filesystem using `memfs` and go back to running stuff in parallel (and give a massive perf boost, tests have a lot of overhead right now).


lishaduck (commented):
Something broke in the rebase, but I'm otherwise satisfied with this iteration, and it seems to fix CI. (Once I fix it the rebase, can you retrigger ci a few more times to double check, @jfmengels?)

jfmengels:
This is a great step forward! Thank you @lishaduck :tada:

jfmengels:
@lishaduck I think the `test-run-record` might be broken, as some snapshots now get deleted. Maybe it's related to GitHub rate limits (but it didn't cause files to be removed). I'll try again later tonight if I get the time.

lishaduck:
> @lishaduck I think the `test-run-record` might be broken, as some snapshots now get deleted. Maybe it's related to GitHub rate limits (but it didn't cause files to be removed). I'll try again later tonight if I get the time.

Hm. Oh, yep. Added this to stop it from destroying my standing with gh while testing:
https://github.com/jfmengels/node-elm-review/blob/7520c6bbf44a3a9225b007c6d39186faeb17a58f/test/run.mjs#L453-L455
It should be `(!SUBCOMMAND || (...))`

Issue #340, opened 2025-02-21
elm-review stalls on Windows after upgrading to 2.13


bdukes:
We're having issues with elm-review after upgrading, I think it's Windows-only. The review application runs but never outputs anything or exits. I can reproduce with a simple example, e.g. https://github.com/bdukes/elm-review-repro (which runs fine on WSL).

I think there are two issues occurring here. The first looks like it has something to do with the tinyglobby npm library. When I duplicate the logic from elm-review in a separate example app, it stalls out with tinyglobby 0.2.10, but finds the files with tinyglobby 0.2.12. However, even on 0.2.12, it does not find the files when it's called from the root directory (i.e. when the file is a child of `cwd`), but only when called from a separate directory. I can try to open an issue on their repo for that (though I'm not 100% sure there's no issue with how elm-review is translating the Windows path to the glob path).

The second issue is that, since elm-review is finding zero files, it is never receiving the `acknowledgeFileReceipt` message and so never knows that it's done. It's not showing the "NO FILES FOUND" message because `options.directoriesToAnalyze.length` is 0, I'm not sure if that's a valid reason to bypass that message in this scenario or not.

Happy to debug and try resolutions.

jfmengels:
@bdukes The fix is released in v2.13.2. I hope this allows you to start the week with a working `elm-review`!

bdukes:
I've upgraded and confirmed that everything is working as expected for us now. Thanks!

PR #344, opened 2025-02-21
Set up CI on Windows

### LINKED ISSUE #343: Add Windows CI
🤦‍♂ See #340.

It'll probably need some normalization, but I'll get it. I don't like breaking my dad :) (or anyone, but especially my dad)

I feel particularly bad because this is the second project migrating to tinyglobby broke, so I should've known better. I thought I had, but clearly not.

lishaduck:
Resolves #343.

~~Should probably fail until #340 gets fixed?~~


jfmengels:
I agree we should run tests on Windows (maybe on Mac as well?), but I imagine it will increase the likelyhood that we get rate limits (which we still keep seeking for some reason). Should we maybe exclude the `template` tests on Windows?

lishaduck:
First off, how did I hit rate limits without running it? Must've still got something wrong with #310. I'll take a look ASAP.
Second off, it luckily looks like most of this so far is newlines. A few path issues _in test code_, but I didn't see any bugs (which are probably in the run tests).

test/run.mjs
```diff
@@ -48,7 +48,13 @@ if (nodeVersion !== expectedVersion) {
  */
 const replaceScript = (data) => {
   const localPath = path.join(__dirname, '..');
-  return data.replace(new RegExp(localPath, 'g'), '<local-path>');
+  return data
+    .replace(
+      // eslint-disable-next-line security/detect-non-literal-regexp -- Test code.
```
lishaduck, line 53: I'm not happy about this either. Perhaps we should only enable security on non-test files?

lishaduck:
Wow, pathe is magic! All tests failing to just one failing because Anonymize needs updates.

test/run.mjs
jfmengels: Depends what other rules `security` holds. I would like it help prevent if people try to do weird code executions for instance. But yeah, a regex like this is fine.

lib/watch.js
```diff
@@ -383,18 +376,15 @@ function createExtraFilesWatcher(options, app, runReview, onError, request) {
   return chokidar
     .watch(
       request.files.map(
-        ({pattern, included}) =>
-          (included ? '' : '!') + OsHelpers.makePathOsAgnostic(pattern)
+        ({pattern, included}) => (included ? '' : '!') + pattern
```
jfmengels, line 379: Does chokidar support Windows-style paths?

lishaduck: ~~No idea. I should check.~~ They normalize it (at least they used to, we're still on v3, so they still should?), so you can use either, but it _should_ return a POSIX path rather than a DOS path.
I had a chat with my dad about the elm-review's handling of drives, which basically ended with `makePathOsAgnostic` uses `/.:/` to strip the drive, when elm-review should keep the drive so that you can execute it across drives (which MS wants to make more common). `pathe` handles proper agnosticism, so the pattern passed in should be correct, but I should also just check that as well.

lishaduck:
@jfmengels, how much do we care about outputting idiomic paths on Windows? Right now, the CLI outputs paths like this: "C:\Somewhere\Foo.txt", but this PR changes that to "C:/Somewhere/Foo.txt", which is _equally valid_, but could be a slight behavioral change (does that hurt the IDE integration?) and might be confusing to newcomers? (But Windows paths are confusing in general, so maybe they should just get over it, see the writeup at https://github.com/jfmengels/node-elm-review/pull/344/commits/b9c822a1113ac15879965b21d53f3db31bd6df9d).

lishaduck:
@jfmengels, can you take a look a the current failure? I'm not really sure why it's different. Is it an prompts thing? I don't think we're doing anything different.

lishaduck:
> Is it an prompts thing?

Yup. This is silly code: https://github.com/terkelg/prompts/blob/e0519913ec4fcc6746bb3d97d8cd0960c3f3ffde/lib/util/figures.js#L24

lishaduck:
> Can Windows really not render the characters by default?

If I'm reading [this old DevBlogs post](https://devblogs.microsoft.com/commandline/windows-command-line-unicode-and-utf-8-output-text-buffer/) correctly, UTF-8 support for Windows Console landed in 2018, so it should be fine by now.

jfmengels:
> how much do we care about outputting idiomic paths on Windows? Right now, the CLI outputs paths like this: "C:\Somewhere\Foo.txt", but this PR changes that to "C:/Somewhere/Foo.txt", which is equally valid, but could be a slight behavioral change (does that hurt the IDE integration?) and might be confusing to newcomers? (But Windows paths are confusing in general, so maybe they should just get over it, see the writeup at https://github.com/jfmengels/node-elm-review/commit/b9c822a1113ac15879965b21d53f3db31bd6df9d).

That's a **lot** of ways to write Windows paths :thinking: 
I honestly don't know. If they're bot h valid, then either one is fine by me, but since I'm not a Windows dev, I'm not adequately positioned to to tell what is best. I would say that it's probably best to change this only it it is nicer for the user in some way, not if it's just for us running it in CI.

lishaduck:
> I would be okay with replacing `prompts` with a custom thing as well, but that's a bigger task 😅

I took a closer look at alternatives. Clack is a very different feel from elm-review's current vibe, I don't think we want to go that way. inquirer also detects windows, but it also supports unicode on modern terminals, so perhaps we could fake the terminal env var in tests to be one that supports unicode? Maybe???
I might need to look into the prior art. Somebody's got to have run into this before.

lishaduck:
Huh. This is weird, I just realized pathe v2 is ESM-only, but we're importing it in CJS without issues (and on v16, not like v23 where that's fine)

lishaduck:
I've been busy with life (still am, hopefully should slow up soon), but have been thinking about this.
1. Let's just find+replace this for now
2. > Clack is a very different feel from elm-review's current vibe

    I've looked into clack some more in general, and while I'd like to use it, it's both ESM-only and would take a lot of effort to make it look like elm-review.

lishaduck:
This PR's diff is going to be such a mess.

lishaduck:
Was looking through the errors and was trying to debug an issue... and it turns out it's 5 years old 😂 
https://github.com/jfmengels/node-elm-review/commit/39ec4e5f5fc36d6c5e3c80d635738771f8132229#diff-9218c1cd9409fa9cb2e842d38893e5e087b82501b9a6accc052ce90c3170cf9b

jfmengels:
This looks great! Great work! Some tests are still failing, for different reasons in both OSes.

lishaduck:
> This looks great! Great work! Some tests are still failing, for different reasons in both OSes.

Yeah, it's not quite done but I think I ported over most of the less invasive changed from the `windows-2` branch + I improved the regexes from this branch & so excepting actual bugs on Windows, I think it shouldn't be too bad to finish.
I'll be on Winter break from school, so I should hopefully be able to find some time to wrap this up soon.

patches/prompts+2.4.2.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/prompts/lib/util/figures.js b/node_modules/prompts/lib/util/figures.js
+index cd31b88..d5c7403 100644
+--- a/node_modules/prompts/lib/util/figures.js
++++ b/node_modules/prompts/lib/util/figures.js
+@@ -28,6 +28,6 @@ const win = {
+   line: '─',	
+   pointer: '>'	
+ };	
+-const figures = process.platform === 'win32' ? win : main;	
++const figures = main;
```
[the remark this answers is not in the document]
lishaduck: > > And this is only used locally, right?
> 
> Yup, it runs in prepare so it only runs when we install, not when users install.

...which turns out to be an issue cuz half of the tests this targets are integration tests that don't pick up our node_modules (on purpose)...
Thus, reverting this, unfortunately. Back to regexes...

lib/error-message.js
```diff
@@ -29,19 +29,19 @@ class CustomError extends Error {
...
+      ),
+      options.forTests
     );
   }
 
-  return Anonymize.pathsAndVersions(options, formatHuman(options.debug, err));
+  return Anonymize.pathsAndVersions(
+    formatHuman(options.debug, err),
+    options.forTests
+  );
```
lishaduck: Note to self: reevalute anonymizing in this function. It's nice to have it, but we end up needing to normalize elsewhere anyway, so perhaps we could simplify this logic by simply removing it.

lishaduck: > I'm guessing we can do a find+replace in the output in a helper from our test suite.

Yeah, we have to do that anyway because this doesn't fully anonymize all error paths, so I figured we might as well drop this (which simplifies some of the windows logic).

PR #345, opened 2025-02-21
Add test for visiting extra files

jfmengels:
I noticed this use-case was missing, and I thought this was relevant to test with #340 / #343 in mind because fetching extra files also uses `tinyglobby`. I am more or less expecting this to be broken on Windows at the moment.


lishaduck:
Realized that we haven't updated snapshots for 2.15.1 (and dependabot isn't able to update them because of config-unparsable-elmjson).

PR #346, opened 2025-02-21
chore: refresh snapshots

lishaduck:


lishaduck:
@jfmengels, can you confirm if the snapshot changes are good?
I'm unsure if they're intended or not.

jfmengels:
I think you removed all the `2.15.1` changes that you had in dedd2d7662a348c6b065a96c25882962d1c1b9a6 and kept the GH rate limit errors, so I think you did the opposite of what you wanted :sweat_smile: 

I'm not sure yet for the suppressed errors. I'll try to look into it tomorrow.

Also, I think `--github-auth` is broken, I can't seem to make it work at all even locally (which would explain why it doesn't work in CI either).

lishaduck:
> I think you removed all the `2.15.1` changes that you had in [dedd2d7](https://github.com/jfmengels/node-elm-review/commit/dedd2d7662a348c6b065a96c25882962d1c1b9a6) and kept the GH rate limit errors, so I think you did the opposite of what you wanted 😅

No, you already upgraded to 2.15.1, I just hadn't pulled, but that commit hit rate limits without you noticing, so this reverts that.

> I'm not sure yet for the suppressed errors. I'll try to look into it tomorrow.

Thanks! ❤️

> Also, I think `--github-auth` is broken, I can't seem to make it work at all even locally (which would explain why it doesn't work in CI either).

Yeah, it didn't seem to work well for me either (I had to manually revert the rate limit issues).

jfmengels:
Ah gotcha. Yeah maybe we should have some additional tests making sure we don't commit GitHub rate limit issues in snapshots. That sounds reasonable 🤔

Let's merge this, I'll look into what was supposed to happen with suppressed errors when I can.

jfmengels:
Digging through the Git history, I think the snapshots were incorrect, and I reverted them back in 5a595f919f1fc2eb284a98e4be10b13c5e2668dd. I don't understand why the snapshots were generated this way :thinking:
Let me know if you see these changing again.

lishaduck:
> Digging through the Git history, I think the snapshots were incorrect, and I reverted them back in [5a595f9](https://github.com/jfmengels/node-elm-review/commit/5a595f919f1fc2eb284a98e4be10b13c5e2668dd). I don't understand why the snapshots were generated this way 🤔 Let me know if you see these changing again.

And CI passed now :)

I took a minute to log the headers getting sent to GH last night, and I figured out why we're still rate limited. Here's the patch:

```diff
diff --git a/lib/remote-template.js b/lib/remote-template.js
index 8f23744d..cb8d33e4 100644
--- a/lib/remote-template.js
+++ b/lib/remote-template.js
@@ -372,7 +372,7 @@ async function makeGitHubApiRequest(options, url, handleNotFound) {
   const parameters = {
     responseType: 'json',
     headers: {
-      Authorization: `BEARER: ${options.gitHubPat}`,
+      Authorization: `BEARER ${options.gitHubPat}`,
       'X-GitHub-Api-Version': '2022-11-28'
     }
   };
 ```

jfmengels:
That works so much better :smile: 
Thanks for finding this, I'll make a PR for it :+1:

PR #350, opened 2025-02-22
Add test about rate limit warnings in tests

jfmengels:
This should help us detect rate limit snapshots. I'm sad I have to do this kind of test, but I made the mistake too many times :sweat_smile:

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "test": "turbo run testing check-engines --continue",
     "test-sync": "npm run jest && npm run test-run",
     "test-run": "(cd test/ && node ./run.mjs)",
+    "test-check-rate-limit-snapshots": "(cd test/ && node ./check-rate-limit-snapshots.mjs)",
     "test-run-record": "(cd test/ && node ./run.mjs record)",
     "tsc": "tsc",
     "tsc-watch": "tsc --watch"
--- a/test/run.mjs
+++ b/test/run.mjs
@@ -29,6 +29,19 @@ const SNAPSHOTS = path.join(__dirname, 'run-snapshots');
 /** @type {string | undefined} */
 const SUBCOMMAND = process.argv[2];
 
+const nodeVersionOutput = await $`node --version`;
+const nodeVersion = nodeVersionOutput.stdout.toString().slice(1).trim();
+const nvmrc = await fsp.readFile('../.nvmrc');
+const expectedVersion = nvmrc.toString().trim();
+
+if (nodeVersion !== expectedVersion) {
+  console.error(`INCORRECT NODE VERSION\n`);
+  console.error(
+    `You are using Node.js version ${nodeVersion} but you should be using ${expectedVersion}`
+  );
+  process.exit(1);
+}
+
 /**
  * @param {string} data
  * @returns {string}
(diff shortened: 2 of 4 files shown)


jfmengels:
I also added a check that we're using the correct Node.js version when running `test-sync`, that will help prevent differences in JSON parsing errors in some of the snapshots.

turbo.json
```diff
@@ -27,6 +27,7 @@
         "tsc",
         "eslint-check",
         "prettier-check",
+        "test-check-rate-limit-snapshots",
         "test-sync"
```
lishaduck, line 31: This'll run them in parallel, which isn't want we want, is it?

package.json
```diff
@@ -56,6 +56,7 @@
     "test": "turbo run testing check-engines --continue",
     "test-sync": "npm run jest && npm run test-run",
     "test-run": "(cd test/ && node ./run.mjs)",
+    "test-check-rate-limit-snapshots": "(cd test/ && node ./check-rate-limit-snapshots.mjs)",
     "test-run-record": "(cd test/ && node ./run.mjs record)",
```
lishaduck, line 60: This should ensure that rate-limited snapshots don't show up.

```suggestion
    "test-run-record": "(cd test/ && node ./run.mjs record && node ./check-rate-limit-snapshots.mjs)",
```

If we also want to run it in CI though, perhaps this would be better?

```suggestion
    "test-check-rate-limit-snapshots": "(cd test/ && node ./check-rate-limit-snapshots.mjs)",
    "test-run-record": "(cd test/ && node ./run.mjs record) && npm run test-check-rate-limit-snapshots",
```

test/run.mjs
```diff
@@ -29,6 +29,19 @@ const SNAPSHOTS = path.join(__dirname, 'run-snapshots');
 /** @type {string | undefined} */
 const SUBCOMMAND = process.argv[2];
 
+const nodeVersionOutput = await $`node --version`;
+const nodeVersion = nodeVersionOutput.stdout.toString().slice(1).trim();
```
lishaduck, line 33: Is there a reason this doesn't work?

```suggestion
const nodeVersion = process.verisons.node;
```

I'd be slightly concerned that some version managers might make shelling out return the wrong answer. Probably not though.

---

I also have a bit of a preference for normalizing the error message like in https://github.com/rtfeldman/node-test-runner/pull/635, but this is certainly better for now.

lishaduck (commented):
Just a question or two.

test/run.mjs
jfmengels: Oh, yeah, maybe we should do that instead :+1:
I'll keep this for now but we can improve this later.

package.json
jfmengels: Sounds great :+1:

turbo.json
jfmengels: I was mostly just trying to get the test in at all tbh.

I agree that it should ideally not be run in parallel. I guess we should run it after `test-sync`  when recording (as you suggested below), and before the tests when not recording (no need to run them all if we know it's incorrect).

Am I correct in thinking this is done just by changing `"test-run"` and `"test-run-record"` in the package.json scripts?

lishaduck: > Am I correct in thinking this is done just by changing "test-run" and "test-run-record" in the package.json scripts?

Yup. I think my a suggestion below should do that.

PR #352, opened 2025-02-23
Report errors if no files were found or if source-directories are empty

jfmengels:
Supersedes #341, by addressing the root problem.

As @lishaduck pointed out in https://github.com/jfmengels/node-elm-review/pull/341#issuecomment-2676439308

> we don't have any tests where there isn't a source dir, do we? I have a sneaking suspicion that the reason it hung was that tinyglobby <0.2.11 would've scanned the entire drive when given an absolute path, but in case I'm wrong there, we should probably make sure elm-review doesn't hang in the actual situation where there aren't any files.

Indeed, when there are no Elm files at all, `elm-review` hangs. I added a test for it, and we're now exiting with an error should that happen. I also added a specific error (and test) for when `source-directories` is empty for some reason.

I also `unique`d source-directories we're running, because there was some duplicates otherwise, and that was one additional `glob` call.

### DIFF
--- a/lib/elm-files.js
+++ b/lib/elm-files.js
@@ -13,6 +13,7 @@ const Debug = require('./debug');
 const ErrorMessage = require('./error-message');
 const FS = require('./fs-wrapper');
 const Hash = require('./hash');
+const {unique} = require('./utils');
 const OsHelpers = require('./os-helpers');
 const elmParser = require('./parse-elm');
 const AppState = require('./state');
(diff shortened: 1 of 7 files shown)


lib/elm-files.js
```diff
@@ -52,6 +53,9 @@ async function getProjectFiles(options, elmSyntaxVersion) {
   const emptyDirectories = filesInDirectories.filter(
     ({files}) => files.length === 0
   );
+  const elmFilesToRead = [
+    ...new Set(filesInDirectories.flatMap((directory) => directory.files))
+  ];
```
lishaduck, line 58: Just noticed that is just unique() inlined. Maybe I'll de-inline it someday. Probably not. :)

jfmengels: Ah right. Yeah I reached out for `unique` after writing this code. I'll make the change.

lishaduck: I was rebasing and this made a conflict, so I fixed it in #344.

jfmengels: Just saw your message. I fixed it in #354, it's on master now :sweat_smile:

lishaduck: Already re-(re-?)rebased

PR #353, opened 2025-02-23
Fix --template used without --github-auth

jfmengels:
If `options.gitHubPat` was undefined, then we sent out `Authorization: BEARER undefined`, which yields a 401 HTTP error. So `--template` was **only** working with `--github-auth`.

I also made a nice error message when we get 401 responses, that should make it more obvious (mostly for us :smile:).


lishaduck:
Amazing 🤦‍♂️

Thanks for catching that.

jfmengels:
Yeah, I was surprised the test suite didn't work without me adding `AUTH_GITHUB` :sweat_smile:

PR #362, opened 2025-04-14
fix: checkout with lf line endings

lishaduck:
Extracted out of #344.

I quit working on that, it's too hacky.
I'm going to split off what I can and then rewrite the rest.


lishaduck:
@jfmengels, this is also ready to merge...

PR #365, opened 2025-05-02
chore: disable caching tests

lishaduck:
Clearly caching isn't working well, I'm going to give a shot at not trying to cache stuff and see if that helps.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,12 +73,7 @@ jobs:
           node_modules-${{ runner.os }}-
 
     - name: Cache turbo build setup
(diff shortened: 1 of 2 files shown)


jfmengels:
@lishaduck Do you think this is still worth merging?
EDIT: You confirmed it was in #362, so merging it :+1:

lishaduck:
> EDIT: You confirmed it was in #362, so merging it 👍

That was outdated...

> @lishaduck Do you think this is still worth merging?

Not really, but I don't think it hurts anything so 👍🏻 from me (for now at least)

PR #367, opened 2025-05-29
Move to `@types/elm`

lishaduck:
Extracted these types out into a package in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/72672. I'm happy if we keep vendoring it, I just needed to test that it worked and figured I might was well file the PR.


jfmengels:
Hey (sorry for the absence)
I'm fine with these changes, but I'd rather they be vendored, so as to reduce the list of dependencies (especially since it's such a small set of type definitions)

lishaduck:
> Hey (sorry for the absence)

Nah, you're fine.
I had to give up on #344 for now anyway b/c WSL 1-to-2 migration broke my path so git bash doesn't work or something weird idk, so I haven't been able to do much over here.

> I'm fine with these changes, but I'd rather they be vendored, so as to reduce the list of dependencies (especially since it's such a small set of type definitions)

Makes sense to me. I was going to suggest we just close it, but on reflection the @types/elm names _are_ more accurate, so it makes sense to me to vendor them.

lishaduck:
I've been busy for a bit myself, but this is finally ready for a re-review.

jfmengels:
Thanks for the better types :relaxed:

PR #374, opened 2025-08-15
Remove check of `hasElmFormatPathFlag` in `ENOENT` error.

realistschuckle:
This suggested change allows the correct error to be printed based on if the person running elm-review has specified the path to the elm-review executable or has not.


jfmengels:
Thank you @realistschuckle :relaxed:

realistschuckle:
@jfmengels `elm-review` is part of my toolchain. I love to help out where I can.

PR #376, opened 2025-08-18
refactor: drop abstractions that make options evaluate strictly

lishaduck:
In order for the next refactor to work, they must be created on-demand.


jfmengels:
Also looks good :+1:

PR #377, opened 2025-08-18
refactor: centralize into functions

lishaduck:
Stop running stuff on module evaluation, instead run it on executing `main()`.

Depends on #380.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -1,19 +1,8 @@
-// TODO(@lishaduck): Create a `setup` function.
-process.title = 'elm-review';
-
-if (!process.argv.includes('--no-color')) {
-  if (process.env.NO_COLOR !== undefined && process.env.NO_COLOR !== '') {
-    process.argv.push('--no-color');
-  } else {
-    // Force `chalk` to add colors by default.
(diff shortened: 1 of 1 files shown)


lib/main.js
```diff
@@ -102,9 +113,11 @@ async function runElmReview() {
 }
 
 /**
+ * @param {Options} options
+ * @param {ErrorHandler} errorHandler
  * @returns {Promise<void | never>}
  */
-async function runElmReviewInWatchMode() {
+async function runElmReviewInWatchMode(options, errorHandler) {
```
lishaduck, line 120: I keep meaning to circle back to this. Unrelated to this change, the recursive nature of this function means I think it leaks memory and could potentially stack-overflow. Or just have bad stacktraces.
Related to this diff, because references to `options` & `errorHandler` now need to passed down and stored on the stack of every recursion, I think this change effectively ⅓s the number of saves elm-review can run it watch mode before crashing. It should still be obscenely large, but still a potential issue?

Then again, I could have just misread the code or perhaps v8 does more optimization to recursion than I think it does.

PR #381, opened 2025-09-05
Start Moving Tests to JS

lishaduck:
~~Blocked on refactoring cwd (#382)~~

~~Depends on #377.~~

### DIFF
--- a/bin/elm-review
+++ b/bin/elm-review
@@ -1,4 +1,4 @@
 #!/usr/bin/env node
 
 // TODO(@lishaduck): With ESM, this should use TLA.
-require('../lib/main')();
+require('../lib/main').main();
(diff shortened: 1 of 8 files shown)


bin/elm-review
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env node
 
 // TODO(@lishaduck): With ESM, this should use TLA.
-require('../lib/main')();
+require('../lib/main').main();
```
lishaduck, line 4: ~~Was just thinking, we should ensure this doesn't break elmls.~~
As far as I can tell, it's just shelling out, correct?
https://github.com/elm-tooling/elm-language-server/blob/d0c9186ec7561abdf7d970be51767304b91d7f5b/src/common/providers/diagnostics/elmReviewDiagnostics.ts#L145-L150

PR #405, opened 2025-12-28
Update snapshots for update-examples-from-preview

jfmengels:
The contents are the same, but we now generate/copy these files using new different permissions, which Git tracks.
This was changed in 67cf975d5fd05a74eea1f497336656351bed50f4


jfmengels:
Haha :sweat_smile: 
Yeah, it looks like our snapshot diffs don't catch it, but when we snapshots, we do get a change. Not great but this should rarely be a problem.

Issue #403, opened 2025-12-28
Unhandled "weird path" in `--config`


miniBill:
```
╰─ bunx elm-review --compiler $(which lamdera) --fix --watch --config ~/src/wolfadex/elm-open-api-cli/cli/review/src/ReviewConfig.elm
-- UNEXPECTED ERROR ------------------------------------------------------------

I ran into an unexpected error. Please open an issue at the following link:
  https://github.com/jfmengels/node-elm-review/issues/new

Please include this error message and as much detail as you can provide. Running
with --debug might give additional information. If you can, please provide a
setup that makes it easy to reproduce the error. That will make it much easier
to fix the issue.

Below is the error that was encountered.
--------------------------------------------------------------------------------
Error: ENOTDIR: not a directory, open '/home/minibill/src/wolfadex/elm-open-api-cli/cli/review/src/ReviewConfig.elm/elm.json'
    at async open (node:internal/fs/promises:641:25)
    at async Object.readFile (node:internal/fs/promises:1245:14)
    at async readFile (/tmp/bunx-1000-elm-review@latest/node_modules/elm-review/lib/fs-wrapper.js:54:16)
    at async Object.readJsonFile (/tmp/bunx-1000-elm-review@latest/node_modules/elm-review/lib/fs-wrapper.js:17:16)
    at async buildLocalProject (/tmp/bunx-1000-elm-review@latest/node_modules/elm-review/lib/build.js:124:5)
    at async Object.build (/tmp/bunx-1000-elm-review@latest/node_modules/elm-review/lib/build.js:110:10)
    at async runElmReviewInWatchMode (/tmp/bunx-1000-elm-review@latest/node_modules/elm-review/lib/main.js:110:5)
    at async main (/tmp/bunx-1000-elm-review@latest/node_modules/elm-review/lib/main.js:264:11)
```

We might want to detect the case where the user passes a `ReviewConfig.elm` instead of just the path. It seems like a reasonable mistake to make. Source: I made it.

jfmengels:
@lishaduck I see your 👀 on this issue. Does that mean you want to tackle this issue? I'm happy to take a look at it but I'd rather avoid duplicate work 🙂

lishaduck:
I was going to, but haven't yet. Feel free to to take it 👍🏻

PR #406, opened 2025-12-28
Improve error message when using --config with a path to a file

jfmengels:
Fixes https://github.com/jfmengels/node-elm-review/issues/403

Preview in image with colors:
image

Also adds a test for a similar untested error message.


jfmengels:
@miniBill what do you think? Would this have helped you? If not, do you perhaps have a suggestion on how to improve the wording?

PR #424, opened 2026-05-11
Remove LICENSE generation and --FOR-TESTS flag

jfmengels:
- When running `elm-review new-package`, the `LICENSE` file will now not be generated automatically. Instead, a manual step has been added in the generated `MAINTENANCE.md` file.
- Anonymize version and path in test code rather than in shipped code
- Remove `--FOR-TESTS` flag and options.forTests.


test/run.mjs
```diff
@@ -14,6 +14,8 @@ import * as process from 'node:process';
 import {fileURLToPath} from 'node:url';
 import {glob} from 'tinyglobby';
 import {$, cd} from 'zx';
+// @ts-expect-error Warning about using "assert" requiring different "module" value in TS. Haven't figured out how.
+import packageJson from '../package.json' assert {type: 'json'};
```
lishaduck, line 18: Assert isn't supported by v16, is it?

lishaduck: Oh, nope, it was backported 👍🏻

lishaduck: I'll take a look tho

lishaduck: Ok, yeah, a ts-expect-error is the best we'll get. TS parses `assert` the same as `with` (with some caveats), and is correctly flagging that `with` is only available in 18+, but we're not using `assert`. Probably a TS bug but there's also almost certainly not appetite for fixing it at this point.

